PR #24, opened 2021-04-25
TCM-105 | Build dummy content screen 

halilyesilyurt:
### Summary

- [x] Completed dummy content screen.
- [x] Added dummy content to test utility.

### Issue

[TCM-105](https://hyfnla.atlassian.net/browse/TCM-105)

### Video/Screenshot

| State Book| State Plain|
|--------|--------|
|content_state_book|content_state_plain|

### DIFF
--- a/app/src/main/java/com/trx/consumer/models/common/ContentModel.kt
+++ b/app/src/main/java/com/trx/consumer/models/common/ContentModel.kt
@@ -1,0 +1,13 @@
+package com.trx.consumer.models.common
+
+import android.os.Parcelable
+import com.trx.consumer.screens.content.ContentViewState
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+class ContentModel(
+    var state: ContentViewState = ContentViewState.PLAIN,
+    var title: String = "",
+    var waiverModel: WaiverModel = WaiverModel(),
+    var body: String = ""
+) : Parcelable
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/trx/consumer/models/common/WaiverModel.kt
```diff
@@ -0,0 +1,17 @@
+package com.trx.consumer.models.common
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+import org.json.JSONObject
+
+@Parcelize
+data class WaiverModel(var covid: String = "", var terms: String = "") : Parcelable {
```
VivekV95, line 8: Think we can remove this as iOS doesn't have it

app/src/main/java/com/trx/consumer/models/common/ContentModel.kt
```diff
@@ -0,0 +1,13 @@
+package com.trx.consumer.models.common
+
+import android.os.Parcelable
+import com.trx.consumer.screens.content.ContentViewState
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+class ContentModel(
+    var state: ContentViewState = ContentViewState.PLAIN,
```
VivekV95, line 9: Let's remove the `state` and `waiverModel` properties to match iOS

app/src/main/java/com/trx/consumer/screens/content/ContentViewState.kt
```diff
@@ -0,0 +1,5 @@
+package com.trx.consumer.screens.content
+
+enum class ContentViewState {
```
VivekV95, line 3: Think we can remove this as iOS doesn't have it

VivekV95 (changes requested):
Few things we can remove that I don't think we'll need in TRX

onodude (approved):

PR #39, opened 2021-05-17
TCM-296 | Connect `Cards`

vespertineSiren:
### Summary

 - Card is being saved in `backendmanager.user()`
 - Add remaining events available on iOS.
 - Use more app specific extensions 

### Issue

[TCM-296](https://hyfnla.atlassian.net/browse/TCM-296)

### Video/Screenshot

https://user-images.githubusercontent.com/27742105/118564110-961b2880-b724-11eb-9fb2-23ddfc03e8b9.mov

https://user-images.githubusercontent.com/27742105/118564117-99aeaf80-b724-11eb-9f0e-80fef10f3336.mov


app/src/main/java/com/trx/consumer/screens/cards/CardsViewModel.kt
```diff
@@ -1,19 +1,64 @@
...
+            val response = backendManager.user()
+            eventShowHud.postValue(false)
+            when {
+                response.isSuccess -> {
+                    cacheManager.user()?.card.let {
+                        model = it
+                        eventLoadViewSuccess.postValue(model)
+                    }
+                }
+                response.statusCode == 404 -> {
```
VivekV95, line 50: This part is a bit confusing to me, I see it's in xpass as well. Why would a 404 indicate success? Wouldn't we want to show failure here?

VivekV95 (changes requested):
Just had a question

app/src/main/java/com/trx/consumer/screens/cards/CardsViewModel.kt
vespertineSiren: Quite frankly, Im not 100% sure either. I know sending a null card will result in the bottom button text to be `Add Payment Method`, hide the card view and show the missing card label. I guess we can make that the default state and then call `eventLoveViewFailure`...

VivekV95: Makes sense to me

onodude: hmm not sure this is correct

onodude: at least for trx should not be needed, xpass probably gave 404 when no cards were found 🤦🏼

devender197:
Few minor changes, 1 and 2 is cosmetic changes and 3 can be functional. With or without it, PR is looking good
1. CardsFragment -> Line 49 -> instead of `it` can we use `model` and inside `let` can we use `safeModel`
2. Line 54, we are using view context and 56 we are using fragment context
3. `setBackAction` can be added in handleLoadViewFailure

app/src/main/java/com/trx/consumer/screens/cards/CardsViewModel.kt
VivekV95: In that case, can we just do 
 ```
 if (response.isSuccess) {
                cacheManager.user()?.card?.let {
                    model = it
                    eventLoadViewSuccess.postValue(model)
                }
            } else {
                eventLoadViewFailure.call()
            }
        }

vespertineSiren: Changed here: https://github.com/hyfn/trx-consumer-android/pull/39/commits/b5e8bac56c654d1a642626a4c576c76789f99a0a

vespertineSiren:
@devender197 
1. ignored. minor change. 
2.  changed here https://github.com/hyfn/trx-consumer-android/pull/39/commits/44185fd40982d32f192d90791144bd3e82069d89
3.  Not changed based on this comment chain: https://github.com/hyfn/trx-consumer-android/pull/39#discussion_r634955323
> Few minor changes, 1 and 2 is cosmetic changes and 3 is more functional. else PR is looking good
> 
> 1. CardsFragment -> Line 49 -> instead of `it` can we use `model` and inside `let` can we use `safeModel`
> 2. Line 54, we are using view context and 56 we are using fragment context
> 3. `setBackAction` is required to add in `handleLoadViewFailure`. Instead of Alert dismiss it should perform `**Take me back**` functionality.

VivekV95 (approved):

onodude (approved):

PR #46, opened 2021-05-19
TCM-268 | Introduce new tab layout

halilyesilyurt:
### Summary

- [x] Created Constant object class.
- [x] Added "isVersion2Enabled" boolean value as default false.
- [x] Changed tabbar home icon.
- [x] Added a new tabbar with expected tabs.

### Issue

[TCM-268](https://hyfnla.atlassian.net/browse/TCM-268)

### Video/Screenshot

"isVersion2Enabled" False

settings

"isVersion2Enabled" True

tabbar_version2enable_true


onodude:
@halilyesilyurt @VivekV95 updated icons on drive once again FYI 🎨

app/src/main/java/com/trx/consumer/constants/Constant.kt
```diff
@@ -0,0 +1,5 @@
+package com.trx.consumer.constants
+
+object Constant {
+    var isVersion2Enabled = false
```
VivekV95, line 4: Let's create a `buildConfigField` in gradle for this instead of introducing a new file

app/src/main/res/menu/second_menu_bottom_nav.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android">
```
VivekV95, line 2: Let's call this menu `menu_bottom_nav` and the other one `menu_bottom_nav_v2` so that they match the versions

app/src/main/java/com/trx/consumer/managers/NavigationManager.kt
```diff
@@ -55,6 +64,12 @@ class NavigationManager {
         graph.startDestination = if (isLoggedIn) R.id.home_fragment else R.id.splash_fragment
         navController.graph = graph
         getTabBar(activity).apply {
+            menu.clear()
```
VivekV95, line 67: Let's revert these lines (we don't need this here because the finalized app will always use `launchSequence`)

app/src/main/java/com/trx/consumer/constants/Constant.kt
VivekV95: Looking again, let's just pretend like v2 doesn't exist, only use the new menu, and remove the checks. Swapping out the tabs and menu should be really simple once it's time to implement v2.

VivekV95: No need to delete the v2 menu file though

halilyesilyurt: completed.

VivekV95 (approved):

onodude (approved):

PR #61, opened 2021-05-24
TCM-373 | Connect workout screen

devender197:
### Summary

- [x] Layout and view name updated to match parity with iOS and different state created
- [x] Booking response model created
- [x] WorkoutViewModel updated 

### Issue

[TCM-373](https://hyfnla.atlassian.net/browse/TCM-373)

### Video

https://user-images.githubusercontent.com/24737954/119398380-53a99c80-bc8c-11eb-9de2-c71237223f46.mov

### Screenshot

Screenshot_1621884631


app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
```diff
@@ -55,6 +66,33 @@ class WorkoutModel(
...
+            when (workoutState) {
+                WorkoutViewState.LIVE -> cancelId = identifier
+                WorkoutViewState.VIRTUAL -> cancelId = sessionId
+                else -> {
+                }
+            }
+            sessionId = identifier.also { identifier = sessionId }
+        }
+
+    val bookViewStatus: BookingState
```
VivekV95, line 87: Let's rename `BookingState` to `BookingViewState` to match iOS

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
```diff
@@ -0,0 +1,46 @@
+package com.trx.consumer.models.responses
+
+import com.trx.consumer.extensions.forEach
+import com.trx.consumer.models.common.WorkoutModel
+import org.json.JSONObject
+
+class BookingResponseModel(val workouts: MutableList<WorkoutModel> = mutableListOf()) {
```
VivekV95, line 7: `workouts` don't need to be mutable

VivekV95, line 7: Let's rename to `BookingsResponseModel` to match iOS

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -50,14 +81,35 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
             btnPrimary.bgColor(model.state.buttonBackgroundColor)
             btnPrimary.textColor(model.state.buttonTitleColor)
             btnPrimary.apply { text = context.getString(model.state.buttonTitle) }
         }
     }
 
+    private val handleTapStartWorkout = Observer<WorkoutModel> { model ->
+        LogManager.log("handleTapStartWorkout")
+        if (model.workoutState == WorkoutViewState.VIDEO) {
+            // NavigationManager.shared.present(this, R.id.player_view) TODO: needs to play video
```
VivekV95, line 95: `Player` is implemented, can check out how to launch it in `TestUtlility`

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
```diff
@@ -55,6 +66,33 @@ class WorkoutModel(
...
+        get() = when (state) {
+            BookingState.BOOKED -> {
+                if (date.elapsedMin() < kMinutesBeforeCanJoin) BookingState.JOIN else BookingState.CANCEL
```
VivekV95, line 90: Exceeds character limit

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
```diff
@@ -0,0 +1,46 @@
...
+
+     val calendarModelVirtual: CalendarModel
+     TODO: need to implement proper logic
+         get() = CalendarModel(CalendarViewState.DISPLAY).apply {
+             listDays = CalendarModel.createListOfDays()
+             listEvents = lstVirtualUpcoming.map { date }
+         }*/
+
+    val lstWorkoutsSorted: List<WorkoutModel>
+        get() = workouts.filter { it.isCanceled }.sortedBy { it.startsAt }.map { it.booking }
```
VivekV95, line 24: Should be filtering for `!it.isCanceled`

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -50,14 +81,35 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
+        }
+
+        if (model.workoutState == WorkoutViewState.LIVE) {
+            NavigationManager.shared.present(this, R.id.live_fragment)
```
VivekV95, line 99: This should be going to `LivePlayer` which hasn't been implemented yet, can add a todo

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -1,21 +1,94 @@
...
+    }
+
+    private fun doLoadWorkout() {
+        if (model.state != BookingState.BOOKED) {
+            viewModelScope.launch {
+                val response = backendManager.bookings()
+                if (response.isSuccess) {
+                    val bookingModel = BookingResponseModel.parse(response.responseString)
+                    bookingModel.lstWorkoutsSorted.firstOrNull { it == model }?.let { model ->
+                        model.state = BookingState.BOOKED
```
VivekV95, line 56: Guessing we'd like to modify our `ViewModel`'s `model` property which it isn't currently doing, maybe we can rename the lambda param to `booking` or something

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -1,21 +1,94 @@
...
+            else -> eventTapProfile.postValue(model.trainer)
+        }
+    }
+
+    fun doTapPrimary() {
+        when (model.workoutState) {
+            WorkoutViewState.VIDEO -> eventTapStartWorkout.postValue(model)
+            WorkoutViewState.LIVE, WorkoutViewState.VIRTUAL -> {
+                if (model.bookViewStatus == BookingState.JOIN) {
+                    eventTapStartWorkout.postValue(model)
```
VivekV95, line 79: We'll want to `return` after this line based on iOS

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -1,21 +1,94 @@
...
+                }
+                viewModelScope.launch {
+                    cacheManager.user()?.card?.let {
```
VivekV95, line 82: Looks like the logic here is reversed compared to iOS, iOS calls the user endpoint if the card is null

VivekV95 (changes requested):
Left some initial thoughts

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
vespertineSiren: There's a typo there. `it.isCanceled` -> it.isCancelled

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
```diff
@@ -0,0 +1,46 @@
...
+    TODO: need to implement proper logic
+        get() =lstWorkoutsSorted.filter { $0.workoutState == . virtual }.filter { $0.date.isLater(than: Date()) }*/
+
+    companion object {
+        fun parse(json: String): BookingResponseModel {
+            return BookingResponseModel().apply {
+                val jsonObject = JSONObject(json)
+                jsonObject.optJSONArray("data")?.forEach { items ->
+                    workouts.add(WorkoutModel.parse(items))
+                }
```
vespertineSiren, line 42: You can probably get the jsonArray and use the `map` extension in file  get this value. This would also remove the need to have a mutableList.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -22,25 +25,53 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
             doLoadView()
         }
 
-        viewBinding.btnBack.setOnClickListener { viewModel.doTapBack() }
+        viewBinding.apply {
+            btnBack.action { viewModel.doTapBack() }
+            btnTrainerProfile.action { viewModel.doTapProfile() }
+            btnPrimary.action { viewModel.doTapPrimary() }
+        }
     }
```
vespertineSiren, line 50: This view initalization stuff is supposed to be before observing the viewmodel events.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -22,25 +25,53 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
+            lblSubtitle.text(model.video.videoDuration)
+
+            viewTrainer.isHidden = false
+            imgTrainerPhoto.load(model.video.trainer.profilePhoto)
+            lblTrainerName.text(model.video.trainer.fullName)
+
+            btnPrimary.apply { text = context.getString(model.bookViewStatus.buttonTitle) }
+            btnPrimary.textColor(model.bookViewStatus.buttonTitleColor)
+            btnPrimary.bgColor(model.bookViewStatus.buttonBackgroundColor)
+        }
```
vespertineSiren, line 72: Consider creating a function for this because of how verbose it is. `textColor` and `bgColor` should be useable in an `apply` block. Refer to `PlansViewHolder`

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -22,25 +25,53 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
         val model = NavigationManager.shared.params(this)
         viewModel.model = if (model is VideoModel) {
             WorkoutModel().apply {
-                mode = ""
-                state = BookingState.VIDEO
+                mode = WorkoutViewState.video
+                state = BookingState.DISABLED
                 video = model
             }
         } else model as WorkoutModel
```
vespertineSiren, line 32: Something about this seems really off to me. Would using `as?` or something cut back on this code?

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewState.kt
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Locale
+
+enum class WorkoutViewState {
+    VIDEO, LIVE, VIRTUAL, UNKNOWN;
+
+    companion object {
+        const val largeGroup = "large-group"
+        const val oneOnOne = "1-on-1"
+        const val smallGroup = "small-group-training"
+        const val video = "video"
```
vespertineSiren, line 12: `const val` should be all caps with underscores as spaces. So this should be something like:

```
const val LARGE_GROUP = "large-group"
const val ONE_ON_ONE = "1-on-1"
const val SMALL_GROUP = "small-group-training"
const val VIDEO = "video"
```

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -22,25 +25,53 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
         viewModel.apply {
             eventTapBack.observe(viewLifecycleOwner, handleTapBack)
+            eventTapProfile.observe(viewLifecycleOwner, handleTapProfile)
+            eventTapBookLive.observe(viewLifecycleOwner, handleTapBookLive)
+            eventTapStartWorkout.observe(viewLifecycleOwner, handleTapStartWorkout)
+
+            eventLoadVideoView.observe(viewLifecycleOwner, handleLoadVideoView)
             eventLoadView.observe(viewLifecycleOwner, handleLoadView)
             doLoadView()
         }
```
vespertineSiren, line 43: Should there be some sort of HUD event here since we are making backend calls?

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -1,21 +1,94 @@
...
+        if (model.state != BookingState.BOOKED) {
+            viewModelScope.launch {
+                val response = backendManager.bookings()
+                if (response.isSuccess) {
+                    val bookingModel = BookingResponseModel.parse(response.responseString)
+                    bookingModel.lstWorkoutsSorted.firstOrNull { it == model }?.let { model ->
+                        model.state = BookingState.BOOKED
+                        model.cancelId = model.cancelId
+                    }
+                    eventLoadWorkoutView.postValue(model)
```
vespertineSiren, line 59: This isn't being handled in the fragment. Remove event entirely or implement in fragment as well.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -22,25 +25,53 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
+            imgTrainerPhoto.load(model.video.trainer.profilePhoto)
+            lblTrainerName.text(model.video.trainer.fullName)
+
+            btnPrimary.apply { text = context.getString(model.bookViewStatus.buttonTitle) }
+            btnPrimary.textColor(model.bookViewStatus.buttonTitleColor)
+            btnPrimary.bgColor(model.bookViewStatus.buttonBackgroundColor)
+        }
+    }
+
     private val handleLoadView = Observer<WorkoutModel> { model ->
```
vespertineSiren, line 75: Consider removing the function body and just adding a log. This is not being called in the viewmodel and is the same contents as `handleLoadVideoView`

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
onodude: lets just change the apply to a normal setting that way this whole block will be consistent

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: Sorry can you please elaborate this. As far as I understand, nothing seems off about this.

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
devender197: I have used the `foreach` function instead of `map`. Hence I need a mutable list.

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
VivekV95, line 90: I'm working with `elapsedMin` and noticed it returns negative values when current time is before the start time. So this condition will equate to true for any time before the start time, which I presume is not what we want. Maybe we should use the absolute value?

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
VivekV95: Using `forEach` is fine, but the implementation shouldn't dictate necessity. If I see an exposed mutable property I'm going to most likely think it's being modified outside of just its creation. Can create the mutable list outside the constructor and expose it as a `List`

VivekV95: Or we can use the `map` extension, I'm fine with either. They both pretty much do the same thing.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: Please look at line number 70 and 71 in `handleLoadVideoView`. it is different in `handleLoadView`

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
devender197: `BookingState` is different than `BookingViewState`.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-ad1c3474489651ab78612134122709e83cc1481a863c936469b91759832da917R53)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-63443031905bf86b4e8fe7d72b97a863621e84b1bdfd870289f177726110fac8R47)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-ad1c3474489651ab78612134122709e83cc1481a863c936469b91759832da917R78)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewState.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-280c9037675cb6eded2f59a3843331989bb472ad78fdaf6d8e4801311ff2fbafR9)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-63443031905bf86b4e8fe7d72b97a863621e84b1bdfd870289f177726110fac8R73)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-63443031905bf86b4e8fe7d72b97a863621e84b1bdfd870289f177726110fac8R35)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-ad1c3474489651ab78612134122709e83cc1481a863c936469b91759832da917R80)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-63443031905bf86b4e8fe7d72b97a863621e84b1bdfd870289f177726110fac8R105)

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-403fadc4d5dc5fea2797de8d8c7e2a6453d5c16a30061a8861aadb271af62f75R91)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-63443031905bf86b4e8fe7d72b97a863621e84b1bdfd870289f177726110fac8R97)

app/src/main/java/com/trx/consumer/models/responses/BookingResponseModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-6134bb6bf5479305875e66945fabdc012aac0ac198051ea7369c33c5c4e83e2dR24)

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
VivekV95: Looks like iOS still has a separate `BookingViewState` it's returning here, let's bring that over

Screen Shot 2021-05-25 at 10 41 50 AM

VivekV95 (changes requested):
Few unaddressed comments and one add-on

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
vespertineSiren: THanks

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
vespertineSiren: thanks

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
vespertineSiren: Thanks

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
vespertineSiren: Was this missed in your commits? I was unable to find.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
vespertineSiren: It can be reduced to one line. Some pseudocode
```if  (model is VideoModel) model.toWorkoutModelExtensionFunction else model as WorkoutModel```

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
devender197: Good catch. But using absolute value will not give us any benefit because after using abs the logic will be true for a given range(eg:  -10 to +10 ) So I modified logic [here](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-403fadc4d5dc5fea2797de8d8c7e2a6453d5c16a30061a8861aadb271af62f75R94) which will do its task.

app/src/main/java/com/trx/consumer/models/common/WorkoutModel.kt
devender197: If you look at `BookingState` and `BookingViewState`, it is kind of similar. Hence I used BookingState. I don't have an issue changes [done](https://github.com/hyfn/trx-consumer-android/pull/61/files#diff-403fadc4d5dc5fea2797de8d8c7e2a6453d5c16a30061a8861aadb271af62f75R90)

devender197:
@VivekV95 @vespertineSirenThank you for review.  I have done all the suggestion. Please let me further changes requried

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
VivekV95: Agreed, I'm not sure we're doing anything like this elsewhere, seems too much like business logic to be done here. A convert function in `VideoModel` would make sense, or we can create a `Params` model for this screen. We can handle it in a followup task.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
VivekV95: Agreed, seems like we'll need a hud here. Can handle it in a followup task

VivekV95 (dismissed):
Approving, a couple things we can handle in a followup along with any additional improvements

Created https://hyfnla.atlassian.net/browse/TCM-395, feel free to add anything that comes to mind

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewState.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/commits/350a346f94c03f4658c880eb7ad73bfd032d22d5)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/61/commits/1961a86689a6dcb38bfeba5db790ce71235b2187)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: I am not agreed. I did this to make parity with iOS. There are many way to concise code. Please check iOS code [here](https://github.com/hyfn/trx-consumer-ios/blob/develop/TRX-Consumer/Screens/Workout/WorkoutViewController.swift#L17)

devender197:
> Approving, a couple things we can handle in a followup along with any additional improvements
> 
> Created https://hyfnla.atlassian.net/browse/TCM-395, feel free to add anything that comes to mind

Handle all the suggestion here. As per my recommendation, no need to create extract task. Thank you

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
vespertineSiren: It just doesn't seem like you're really leveraging kotlin by leaving it this way.

onodude (approved):
thank you!

PR #63, opened 2021-05-25
TCM-361 | Connect selected filters to backend

devender197:
### Summary

- [x] Connect the selected filters to the backend.
- [x] Disabled multi select will be single selection in filter option screen.
- [x] Selected filters on discover screen changed to different background and text color

### Issue

[TCM-361](https://hyfnla.atlassian.net/browse/TCM-361)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/119578361-38fd2380-bd71-11eb-8976-81b0890ebba3.mov

### Filter with video url on applying filters

https://user-images.githubusercontent.com/24737954/119580968-71533080-bd76-11eb-8e44-4475c4ea9707.mov


onodude (changes requested):
@devender197 pushed some changes. there is one more item to be done:
- when the filter is not selected on the discover screen the view group should have a white background and a border

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
```diff
@@ -2,10 +2,41 @@ package com.trx.consumer.models.params
...
+                        values = mutableListOf<FilterOptionsModel>().apply {
+                            it.values.forEach { option ->
+                                add(FilterOptionsModel(option.value, option.isSelected))
+                            }
+                        }
+                    )
+                )
+            }
+        }
+        return this.copy(selectedModel = FilterModel(), lstFilters = list)
```
VivekV95, line 40: Is the `copy` needed if we're setting both properties? Could just return a new instance

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
```diff
@@ -2,10 +2,41 @@ package com.trx.consumer.models.params
...
+                    val selectedFilter = filter.values.filter { it.isSelected }
+                    if (selectedFilter.isNotEmpty())
+                        put(filter.title, selectedFilter.map { it.value })
+                }
+            }
+            return if (params.isNotEmpty()) params else null
+        }
+
+    fun copyModel(): FilterParamsModel {
+        val list = mutableListOf<FilterModel>().apply {
```
VivekV95, line 26: Let's try to use `map` wherever possible instead of the Java-style loop though + add, will usually save a decent amount of code:
```
  val list = lstFilters.map {
            FilterModel(
                title = it.title,
                values = it.values.map { option ->
                    FilterOptionsModel(option.value, option.isSelected)
                }
            )
        }

File: app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt [hunk 3]
```diff
@@ -2,10 +2,41 @@ package com.trx.consumer.models.params
...
+import com.trx.consumer.models.common.FilterOptionsModel
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class FilterParamsModel(
     var selectedModel: FilterModel = FilterModel(),
     var lstFilters: List<FilterModel> = listOf()
-) : Parcelable
+) : Parcelable {
+    val selectedFilterParams: HashMap<String, Any>?
```
VivekV95, line 13: Needs an empty line here

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
```diff
@@ -28,6 +28,10 @@ class DiscoverFragment : BaseFragment(R.layout.fragment_discover) {
     private var currentState = DiscoverViewState.WORKOUT
 
     override fun bind() {
+        NavigationManager.shared.params(this)?.let { params ->
+            if (params is FilterParamsModel)
```
VivekV95, line 32: Can fit on one line

VivekV95, line 32: Can fit on one line

app/src/main/java/com/trx/consumer/screens/discover/discoverfilter/DiscoverFilterViewHolder.kt
```diff
@@ -14,6 +14,9 @@ class DiscoverFilterViewHolder(view: View) : CommonViewHolder(view) {
 
     fun setup(item: FilterModel, listener: DiscoverFilterListener) {
         lblFilter.text = item.title
+        val isSelected = item.values.find { it.isSelected }
+        lblFilter.textColor(if (isSelected != null) R.color.white else R.color.black)
+        viewMain.bgColor(if (isSelected != null) R.color.black else R.color.greyLightExtra)
```
VivekV95, line 19: Based on the design specs, looks like we want a border instead of background color
Screen Shot 2021-05-24 at 11 09 01 PM

VivekV95: When it's not selected at least

VivekV95 (changes requested):
Left a few thoughts. Also, is the endpoint fully functional with the params? Not seeing any difference in the data returned with filters selected (on prod):

https://user-images.githubusercontent.com/39810816/119438562-1ffb6080-bce6-11eb-8ebc-85dd3bf5dabb.mov

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
devender197: Yes, we can return the new instance also. But when we change the value of any variable of a particular object and use copy method. it will generate the new instance of that class. I checked it. In this case, both is doing same thing.

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
devender197: Doing the same thing.  `foreach` and `map` is both part of Kotlin extension. I feel that using foreach increase code readability.

app/src/main/java/com/trx/consumer/screens/discover/discoverfilter/DiscoverFilterViewHolder.kt
devender197: will be address [here](https://github.com/hyfn/trx-consumer-android/pull/63#pullrequestreview-667334829)

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/63/commits/f873a0ac24de2a99860c0bd90c24186e4f763c71#diff-4643fe9ce6a7797ec074da0c8e709df9eaf34d078db4fc2602e0335b6a9eefc3R13)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/63/commits/f873a0ac24de2a99860c0bd90c24186e4f763c71#diff-b8b5acbef5ed93a44a1868dafc1fb0a8030303ec8efaac50ea795075a767ed99R32)

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
VivekV95: I'm not sure I understand, `copy` does things in the background that we're not using as we're overriding both properties, so why call it?

VivekV95: Come to think of it, does this even need to be a data class?

app/src/main/java/com/trx/consumer/models/params/FilterParamsModel.kt
VivekV95: Readability is a fair argument, but in this particular case I don't think it's more readable as we have 6 levels of indentation for a couple of basic operations

devender197:
> @devender197 pushed some changes. there is one more item to be done:
> 
> * when the filter is not selected on the discover screen the view group should have a white background and a greyLight border with 1dp

[Done](https://github.com/hyfn/trx-consumer-android/pull/63/files#diff-db9f0088ec3fd8b6d7d2b774b7e50ff443cda3d815982aa3646720e6036f96bdR20)

onodude (approved):
thank you!

PR #69, opened 2021-05-26
TCM-381 | Button State Login

vespertineSiren:
### Summary

- Enable/disable button based on contents in InputView
- Create function that enables button state and switches color based on two static values. 

### Issue

[TCM-381](https://hyfnla.atlassian.net/browse/TCM-381)

### Video/Screenshot

| Initial | Filled |
|----|----|
|initial|filled|

https://user-images.githubusercontent.com/27742105/119737184-b7ba9500-be33-11eb-9dac-f5a2c0b5b590.mov

### DIFF
--- a/app/src/main/java/com/trx/consumer/common/CommonButton.kt
+++ b/app/src/main/java/com/trx/consumer/common/CommonButton.kt
@@ -6,7 +6,6 @@ import android.text.SpannableStringBuilder
 import android.util.AttributeSet
 import androidx.appcompat.widget.AppCompatButton
 import androidx.core.content.ContextCompat
-import com.trx.consumer.R
 import com.trx.consumer.common.shapeable.CommonShapeable
 import com.trx.consumer.common.shapeable.CommonShapeableHandler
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/trx/consumer/common/CommonButton.kt
```diff
@@ -44,4 +45,11 @@ class CommonButton @JvmOverloads constructor(
     fun textColor(color: Int) {
         this.setTextColor(ContextCompat.getColor(this.context, color))
     }
+
+    fun setPrimaryEnabled(isEnabled: Boolean) {
+        if (isEnabled != isEnabled()) {
+            bgColor(if (isEnabled) R.color.black else R.color.greyDark)
```
VivekV95, line 51: Not sure we want specific colors built into a reusable view, would be better as an extension function

vespertineSiren: I thought it through that way but I was recommended to code it this way initially. Though that is a good suggestion for a refactor if given the greenlight in the future.

VivekV95 (changes requested):
Just one thought

VivekV95 (approved):

app/src/main/java/com/trx/consumer/common/CommonButton.kt
VivekV95: Confirmed with @onodude, let's make this an extension

VivekV95 (changes requested):
Just the extension

app/src/main/java/com/trx/consumer/common/CommonButton.kt
vespertineSiren: Cool will inline.

onodude (approved):
thank you!

PR #78, opened 2021-05-28
TCM-419 | Add signUpUserType to Register params

vespertineSiren:
### Summary

- Add `signUpUserType` to the params in Register. This will allow users to register on android. 

### Issue

[TCM-491](https://hyfnla.atlassian.net/browse/TCM-419)

### Video/Screenshot

No video but register will bring you to update when signing up with an unclaimed email, valid password, and agreeing to terms. The user will receive a verification message with the email they registered with. 

Screen Shot 2021-05-27 at 11 21 53 PM


onodude (approved):
thanks!

PR #89, opened 2021-05-30
TCM-368 | Polish plans screen

vespertineSiren:
### Summary

- Follow [figma](https://www.figma.com/file/LXa0oBt9aSNOk0YMAeSM8Q/TRX-Consumer?node-id=1703%3A2) for `Plans` from ticket. 
- Make sure the data is connected. 

### Issue

[TCM-368](https://hyfnla.atlassian.net/browse/TCM-368)

### Video/Screenshot

Prod Login: `ashley+prod3@commaeight.com`

https://user-images.githubusercontent.com/27742105/120122997-5bbc7d00-c161-11eb-8735-31e929e2c817.mov

| No Plan | Plan | 
|----|----|
|Screenshot_1622427563|Screenshot_1622427398|


app/src/main/java/com/trx/consumer/screens/plans/PlansFragment.kt
```diff
@@ -62,6 +62,7 @@ class PlansFragment : BaseFragment(R.layout.fragment_plans) {
         LogManager.log("handleLoadCanCancel")
         viewBinding.apply {
             btnCancel.isHidden = !value
+            btnCancel.action { viewModel.doTapUnsubscribe() }
```
VivekV95, line 65: Should we show the cancel alert here?

app/src/main/java/com/trx/consumer/models/responses/PlansResponseModel.kt
```diff
@@ -11,15 +11,19 @@ class PlansResponseModel(private var plan: PlansModel = PlansModel()) {
...
                     list.removeAt(index).let { planModel ->
+                        list.clear()
                         planModel.primaryState = PlansViewState.CURRENT
                         list.add(0, planModel)
                     }
                 }
             }
             list
-        } ?: plan.plans
+        } ?: listOf()
```
VivekV95, line 26: Screen Shot 2021-05-30 at 8 53 53 PM

Looks like we should be doing `userModel?.planText?.let { }` here in `PlansViewModel`, which means that `planText` can be passed in as non-null and hence no need to the null handling

VivekV95 (changes requested):
Couple thoughts

app/src/main/java/com/trx/consumer/models/responses/PlansResponseModel.kt
vespertineSiren: Here we are: https://github.com/hyfn/trx-consumer-android/pull/89/commits/38a446e8b75742298e909262db9a924e26fb7499

app/src/main/java/com/trx/consumer/screens/plans/PlansFragment.kt
vespertineSiren: I have no preference either way

VivekV95: Think that'd be safer and more consistent with our usual transaction stuff, user could easily hit the button accidentally. Not sure what iOS is currently doing

vespertineSiren: Theyre doing this weird thing that I was sure on how to implement. 
`self.viewMain.btnCancelPlan.action{ self.handleLoadCancelPlan(CacheManager.shared.user?.plan) }`

VivekV95: Yeah looks like that ends up showing the cancel alert, we'll just have to create a `ViewModel` function like `doLoadCancelPlan` or something

vespertineSiren: Here we go: https://github.com/hyfn/trx-consumer-android/pull/89/commits/335bdeac784a8e390ed7094adcd0de17c63bf25f

onodude (approved):
thank you!

PR #96, opened 2021-06-01
TCM-463 |  Improve discover filters flow

devender197:
### Summary

- [x] Navigate to workouts tab if filters are selected.
- [x] Clear the filters and decrease the opacity of the filters component to 0.3 when collections and programs tabs are selected

### Issue

[TCM-463](https://hyfnla.atlassian.net/browse/TCM-463)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/120390946-33f52280-c2e3-11eb-8047-288f090c69a2.mov


app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
```diff
@@ -50,13 +50,13 @@ class DiscoverFragment : BaseFragment(R.layout.fragment_discover) {
...
 
-            btnWorkouts.action { viewModel.doLoadWorkouts() }
+            btnWorkouts.action { viewModel.doLoadVideos() }
             btnCollections.action { viewModel.doLoadCollections() }
             btnPrograms.action { viewModel.doLoadPrograms() }
             btnFilter.action { viewModel.doTapFilter() }
         }
 
-        viewModel.doLoadView()
+        viewModel.doLoadVideos()
```
VivekV95, line 59: `doLoadView` is our common pattern, can call `doLoadVideos` within `doLoadView`

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -90,6 +108,10 @@ class DiscoverViewModel @ViewModelInject constructor(
 
     override fun doTapDiscoverFilter(filter: FilterModel) {
         params.selectedModel = filter
-        eventTapDiscoverFilter.postValue(params)
+        if (isFilterEnable) eventTapDiscoverFilter.postValue(params)
+    }
+
+    fun setFilterClick(isClickable: Boolean) {
```
VivekV95, line 114: Let's used either `enabled` or `clickable` across for naming instead of both, bit confusing to follow otherwise

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -23,6 +23,7 @@ class DiscoverViewModel @ViewModelInject constructor(
     var programs: List<VideosModel> = listOf()
     var params: FilterParamsModel = FilterParamsModel()
     var filters: List<FilterModel> = listOf()
+    var isFilterEnable: Boolean = true
```
VivekV95, line 26: `isFilterEnabled`

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -60,24 +62,40 @@ class DiscoverViewModel @ViewModelInject constructor(
         }
     }
 
+    private suspend fun doLoadFilteredWorkout(paramsToSend: HashMap<String, Any>): List<VideoModel> {
```
VivekV95, line 65: `doLoadFilteredWorkouts`

VivekV95 (changes requested):
Logic looks good, just some naming stuff

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
VivekV95: Can also make this `private set` otherwise the setter becomes redundant

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/files#diff-7d6537adf3387efed84a304a14a1f45a83ad7b40211b6512c3ed144eccee7e55R114)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/files#diff-7d6537adf3387efed84a304a14a1f45a83ad7b40211b6512c3ed144eccee7e55R26)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/files#diff-7d6537adf3387efed84a304a14a1f45a83ad7b40211b6512c3ed144eccee7e55R65)

devender197:
@VivekV95 Thank you for review. I have implemented the requested change. Please let me know for further enhancements. Thank you.

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
vespertineSiren: I second this. Should be included all the same for pattern consistency.

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
```diff
@@ -146,4 +149,9 @@ class DiscoverFragment : BaseFragment(R.layout.fragment_discover) {
             }
         }
     }
+
+    private fun handleFilterClick(isClickable: Boolean) {
```
vespertineSiren, line 153: I'm not sure I agree with the naming of this function. The `handle` prefix is usually reserved for `Observer<T>`. Maybe do something like `loadFilterClick` or something along those lines.

vespertineSiren: Going down to line 55, we're also calling a function in the viewmodel.... wouldn't an `event` for this make more sense? I dont see many handlers in the app call a viewmodel function directly.

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -19,10 +19,12 @@ class DiscoverViewModel @ViewModelInject constructor(
 ) : BaseViewModel(), DiscoverListener, DiscoverFilterListener {
 
     var workouts: List<VideoModel> = listOf()
+    var filteredWorkouts: List<VideoModel> = listOf()
     var collections: List<VideosModel> = listOf()
     var programs: List<VideosModel> = listOf()
     var params: FilterParamsModel = FilterParamsModel()
     var filters: List<FilterModel> = listOf()
+    private var isFilterEnabled: Boolean = true
```
vespertineSiren, line 27: I'm not a big fan of how this variable is being used. Can't you enable/disable the views that are associated with calling `doTapFilter` and `doTapDiscoverFilter` instead of using that variable for checks?

app/src/main/res/layout/fragment_discover.xml
```diff
@@ -118,15 +118,14 @@
...
-            android:layout_marginStart="5dp"
+            android:layout_marginStart="10dp"
             android:orientation="horizontal"
             app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
             app:layout_constraintBottom_toBottomOf="parent"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toEndOf="@id/btnFilter"
             app:layout_constraintTop_toTopOf="parent"
             tools:listitem="@layout/row_discover_filter" />
-
```
vespertineSiren, line 129: Why was this removed?

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -60,24 +63,41 @@ class DiscoverViewModel @ViewModelInject constructor(
...
+        resetFilters()
         eventLoadCollections.postValue(collections)
     }
 
     fun doLoadPrograms() {
+        resetFilters()
         eventLoadPrograms.postValue(programs)
     }
 
+    private fun resetFilters() {
```
vespertineSiren, line 89: `doResetFilters()`?

vespertineSiren (changes requested):
Few thoughts. 

Please add `//regions [Region Name]` and `//endregion` for easier organization.

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -19,10 +19,12 @@ class DiscoverViewModel @ViewModelInject constructor(
 ) : BaseViewModel(), DiscoverListener, DiscoverFilterListener {
 
     var workouts: List<VideoModel> = listOf()
+    var filteredWorkouts: List<VideoModel> = listOf()
```
VivekV95, line 22: We're not storing these based on our discussion, just sending them

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
VivekV95: To add on, do we want filters to be scrollable but not clickable, or disable all interaction?

VivekV95: Confirmed with @onodude, we'll want to just disable all interaction with filters, so shouldn't be able to click or scroll when on collections or programs

onodude: yup all the above are correct. we can remove the variable, lets have a method called `setFiltersEnabled` in the fragment which we can just call in the load workouts collections and programs handlers

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: We have reset functionin FilterOptionsViewModel. So I think name is fine. @onodude please let me know if it is required

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: This variable is not created to store the list. It is part of logic. Just consider situation, user want to selected few filter and now we need to get a filtered list from server by calling video(params) and wants to get workout, collection and program by calling videos() which is required. After getting this I want to call the doWorkout only once. If I separate both the call individually user can see a flick first workout list from video and then filtered workout. For any further discussion we can talk on call or group channel.

app/src/main/res/layout/fragment_discover.xml
devender197: [Changes reverted](https://github.com/hyfn/trx-consumer-android/pull/96/commits/8b23f6d62a3052f6d9c51f155322c74f1f136e48#diff-34fc9a1ee677f41779300226835d10828342101e207577c98de99567a74a3fb8R129)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/commits/8b23f6d62a3052f6d9c51f155322c74f1f136e48#diff-b8b5acbef5ed93a44a1868dafc1fb0a8030303ec8efaac50ea795075a767ed99R154)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: RecyclerView scroll is disabled [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/8b23f6d62a3052f6d9c51f155322c74f1f136e48#diff-b8b5acbef5ed93a44a1868dafc1fb0a8030303ec8efaac50ea795075a767ed99R157)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
VivekV95: Makes sense, but if we have the ability to preserve the filtered workouts, not sure why resetting the filters when another tab is tapped is a requirement, will discuss tomorrow

devender197:
@onodude @VivekV95 @vespertineSiren. Thank you for review. I have done the requested change in team meeting. Please review it and let me if any enhancement is required.

1. Removed isFilteredEnabled and FilteredWorkouts from viewmodel [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/60162cbc1a8ad0eb2161607b58c74e51426213ee#diff-7d6537adf3387efed84a304a14a1f45a83ad7b40211b6512c3ed144eccee7e55L27) and [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/5b8d394ed209942ad00562970093ca4031b8998f#diff-7d6537adf3387efed84a304a14a1f45a83ad7b40211b6512c3ed144eccee7e55L22)
2. Disable the scroll and child click listeners for recyclerview [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/60162cbc1a8ad0eb2161607b58c74e51426213ee#diff-fab3d51859943ce286edb4d9ad505f98945651bf69c0dcbfc34820adc14a5383R24) and [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/60162cbc1a8ad0eb2161607b58c74e51426213ee#diff-b8b5acbef5ed93a44a1868dafc1fb0a8030303ec8efaac50ea795075a767ed99R156)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/commits/38f58431a7b17b8fe4178b0059e0102d556e9341)

devender197:
> Few thoughts.
> 
> Please add `//regions [Region Name]` and `//endregion` for easier organization.

@vespertineSiren Thank you . Region added [here](https://github.com/hyfn/trx-consumer-android/pull/96/commits/0a1545a7028b9e3a4e84bbbc6ee1a903e2d77e26)

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
devender197: [Removed](https://github.com/hyfn/trx-consumer-android/pull/96/commits/5b8d394ed209942ad00562970093ca4031b8998f)

app/src/main/java/com/trx/consumer/common/CommonRecyclerView.kt
```diff
@@ -14,13 +14,15 @@ class CommonRecyclerView @JvmOverloads constructor(
     init {
         addOnItemTouchListener(object : SimpleOnItemTouchListener() {
             override fun onInterceptTouchEvent(rv: RecyclerView, e: MotionEvent): Boolean {
-                return if (isScrollable) false else rv.scrollState == SCROLL_STATE_DRAGGING
+                return !isUserInteractionEnabled
```
VivekV95, line 17: We'll be bringing over these changes to Xpass and other projects so we'll still want to account for the old disabling scrolling behavior, with this change disabling scrolling will have no effect. Flow should go like this:
- if `isUserInteractionEnabled` is false, then just return false
- otherwise, return whatever the value of `isScrollable` is

app/src/main/java/com/trx/consumer/extensions/ViewExtensions.kt
```diff
@@ -43,6 +45,19 @@ fun View.dimensions(width: Float? = null, height: Float? = null) {
     }
 }
 
+fun View.isEnabled(enabled: Boolean) {
```
VivekV95, line 48: This seems way out of scope, we just wanted to disable `CommonRecyclerView` interaction, there's no guarantee this is the behavior we'll want outside of the `Discover` screen and it likely won't be

devender197: It will useful in future if we want disable any view and all the childview.

app/src/main/java/com/trx/consumer/common/CommonRecyclerView.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/96/commits/715719a9780797214db90dda3a8e165710d9236c)

app/src/main/java/com/trx/consumer/common/CommonRecyclerView.kt
```diff
@@ -14,13 +14,16 @@ class CommonRecyclerView @JvmOverloads constructor(
     init {
         addOnItemTouchListener(object : SimpleOnItemTouchListener() {
             override fun onInterceptTouchEvent(rv: RecyclerView, e: MotionEvent): Boolean {
-                return if (isScrollable) false else rv.scrollState == SCROLL_STATE_DRAGGING
+                return if (!isUserInteractionEnabled) true
```
onodude, line 17: `return if (!isUserInteractionEnabled) true`
instead
`return if (isUserInteractionEnabled) false`

app/src/main/java/com/trx/consumer/extensions/ViewExtensions.kt
onodude: could you please remove the extension please is just used in one place and implement the method we discussed

onodude (changes requested):
thanks, left a couple of notes

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
```diff
@@ -118,6 +129,7 @@ class DiscoverFragment : BaseFragment(R.layout.fragment_discover) {
     private fun loadPrograms(videos: List<VideosModel>) {
         val state = DiscoverViewState.PROGRAMS
         loadTabs(state)
+        setFilterEnabled(false)
```
onodude, line 132: i guess not needed since its also called in `loadTabs`

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -36,15 +39,22 @@ class DiscoverViewModel @ViewModelInject constructor(
     val eventTapFilter = CommonLiveEvent<FilterParamsModel>()
     val eventTapDiscoverFilter = CommonLiveEvent<FilterParamsModel>()
 
+    //endregion
```
onodude, line 42: thank you for the regions

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
onodude: does not have a event flow attached so can be resetFilters imo

app/src/main/java/com/trx/consumer/common/CommonRecyclerView.kt
onodude: sorry this breaks the logic, nvm

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -56,22 +66,38 @@ class DiscoverViewModel @ViewModelInject constructor(
...
+    private suspend fun doLoadFilteredWorkouts(paramsToSend: HashMap<String, Any>) {
+        val response = backendManager.videos(paramsToSend)
+        if (response.isSuccess) {
+            val model = VideosResponseModel.parse(response.responseString)
+            eventLoadWorkouts.postValue(model.results)
+        } else doLoadWorkouts()
+    }
+
     fun doLoadWorkouts() {
+        eventLoadFilters.postValue(filters)
```
VivekV95, line 82: I _think_ this can be removed as filters are posted in the `doLoadVideos` flow and we don't need to do it again, please correct me if I'm wrong though

VivekV95: Nvm I think we do need it as selected filters gets reset

app/src/main/java/com/trx/consumer/extensions/ViewExtensions.kt
devender197: [Reomved](https://github.com/hyfn/trx-consumer-android/pull/96/commits/39ddaa9e622703d41cb9f99458b797e678f398cc#diff-1486743caa850524ef37fe308ca92963c99f21d27c84527c5b4bc7175e72e969L48)

app/src/main/java/com/trx/consumer/common/CommonRecyclerView.kt
devender197: Done

app/src/main/java/com/trx/consumer/screens/discover/DiscoverFragment.kt
devender197: [Removed](https://github.com/hyfn/trx-consumer-android/pull/96/commits/8658b806b7d1fd8c9d35ab0fbd840a33815f9738#diff-b8b5acbef5ed93a44a1868dafc1fb0a8030303ec8efaac50ea795075a767ed99L131)

onodude (approved):
thank you!

PR #99, opened 2021-06-02
TCM-467 | Replace LoadingHUD's JSON

vespertineSiren:
### Summary

- Change to JSON, Small change to XML. 

### Issue

[TCM-467](https://hyfnla.atlassian.net/browse/TCM-467)

### Video/Screenshot

NEW - No white box

https://user-images.githubusercontent.com/27742105/120576196-3b95f380-c3d7-11eb-94c8-b79f9333dbec.mov

OLD

https://user-images.githubusercontent.com/27742105/120539901-f05eef00-c39c-11eb-9b20-612e79e429f9.mov


app/src/main/res/layout/layout_hud.xml
```diff
@@ -19,17 +20,19 @@
         app:cornerRadius="15dp">
 
         <com.trx.consumer.common.CommonImageView
+            android:id="@+id/imgSquare"
```
VivekV95, line 23: Are we still keeping the white container with the new animation? Looks a bit odd

vespertineSiren: Yeah it was on the iOS PR.

app/src/main/res/layout/layout_hud.xml
```diff
@@ -19,17 +20,19 @@
...
+            android:layout_height="match_parent"
             android:alpha="0.5"
             android:background="@color/white" />
 
         <com.airbnb.lottie.LottieAnimationView
             android:id="@+id/animationView"
             style="@style/MatchParentEdgesViewStyle"
-            android:layout_width="80dp"
-            android:layout_height="80dp"
+            android:layout_width="match_parent"
```
VivekV95, line 33: can just have `style="@style/MatchParentDimensionsViewStyle"` instead of the current style + width + height

VivekV95 (changes requested):
Couple thoughts

onodude:
@vespertineSiren @VivekV95 new request came in to remove the box. so we should just have the full bg with a dim and then the animation

VivekV95 (changes requested):
There are a couple lines within `CommonHudView` that we can un-comment that I think would be within scope, looking good otherwise

VivekV95 (approved):

onodude (approved):
looks great @vespertineSiren

PR #102, opened 2021-06-03
TCM-189 | Check subscription before playing video workouts

VivekV95:
## Description
Check to see if user has "ondemand" subscription before viewing video

### Summary

- [x] Check for subscription before starting video workout
- [x] Direct user to `Subscriptions` screen if user doesn't have one and wants to subscribe
- [x] Display error on `purchases` error  

### Issue

[TCM-189](https://hyfnla.atlassian.net/browse/TCM-189)

### Video/Screenshot

[With valid subscription](https://drive.google.com/file/d/1B_Fke4bHrt8d00voUaUzRE2483CcBUxW/view?usp=sharing)
[No valid subscription](https://drive.google.com/file/d/1odc4gK7V9PMpV4TGR8KaFLwXmBU-yij7/view?usp=sharing)
[Response failure](https://drive.google.com/file/d/1nNVwWfXfWgzGKkOSuWmRRHeljWnuXePy/view?usp=sharing)


devender197 (dismissed):
Looking good to me..

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -76,23 +81,50 @@ class WorkoutViewModel @ViewModelInject constructor(
 
     fun doTapPrimary() {
         when (model.workoutState) {
-            WorkoutViewState.VIDEO -> eventTapStartWorkout.postValue(model)
-            WorkoutViewState.LIVE, WorkoutViewState.VIRTUAL -> {
-                if (model.bookViewStatus == BookingViewState.JOIN) {
+            WorkoutViewState.VIDEO -> {
+                if (!isVersion1Enabled) {
```
devender197, line 85: Please consider using `if else` instead of `return` or if we can avoid `!` by using positive case first.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
```diff
@@ -76,23 +81,50 @@ class WorkoutViewModel @ViewModelInject constructor(
...
+                        it.iapPackage.product.sku == entitlement.productIdentifier
+                    }?.let { match ->
+                        LogManager.log("Matched purchased package ${match.iapPackage.product.sku}")
+                        eventTapStartWorkout.postValue(model)
+                    } ?: eventLoadError.postValue("No matching package")
+                } ?: eventLoadSubscribePrompt.call()
+            } else {
+                eventLoadError.postValue("Could not load purchases")
+            }
+        }.invokeOnCompletion { eventShowHud.postValue(false) }
```
devender197, line 114: If it is not a special case, can we use `eventShowHud` directly in end of scope without `invokeOnCompletion`

devender197:
Seems good to me, added two suggestions. Thank you.

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
VivekV95: [Done](https://github.com/hyfn/trx-consumer-android/pull/102/files#diff-ad1c3474489651ab78612134122709e83cc1481a863c936469b91759832da917R85)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutViewModel.kt
VivekV95: Don't remember a special case here, [Done](https://github.com/hyfn/trx-consumer-android/pull/102/files#diff-ad1c3474489651ab78612134122709e83cc1481a863c936469b91759832da917R110)

onodude:
@VivekV95 i guess we will do a fresh PR after memberships, thoughts?

onodude:
thank you!

PR #104, opened 2021-06-04
TCM-477 | For You section

vespertineSiren:
### Summary

Add space if `For You` section is empty. 

### Issue

[TCM-477](https://hyfnla.atlassian.net/browse/TCM-477)

### Video/Screenshot

Screenshot_1623096322

https://user-images.githubusercontent.com/27742105/121080434-457d7500-c790-11eb-8282-6a3ba94b2403.mov

https://user-images.githubusercontent.com/27742105/121080443-47dfcf00-c790-11eb-956c-117cc776b769.mov

### DIFF
--- a/app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
+++ b/app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
@@ -146,7 +146,7 @@ class HomeFragment : BaseFragment(R.layout.fragment_home) {
         promos.isEmpty().let { isEmpty ->
             if (!isEmpty) promoAdapter.update(promos)
             viewBinding.apply {
-                imgLineOnDemand.isHidden = isEmpty
+                imgLineOnDemand.isGone = isEmpty
                 viewPromos.viewMain.isGone = isEmpty
                 viewPromos.lblTitle.text = getString(R.string.promos_top_title_label)
             }
--- a/app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
+++ b/app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
@@ -1,6 +1,6 @@
 package com.trx.consumer.screens.home
 
-import androidx.core.view.isInvisible
+import androidx.core.view.isGone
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
```diff
@@ -143,12 +144,10 @@ class HomeFragment : BaseFragment(R.layout.fragment_home) {
     }
 
     private fun loadPromos(promos: List<PromoModel>) {
-        val hide = promos.isEmpty()
-        promoAdapter.update(promos)
+        if (promos.isNotEmpty()) promoAdapter.update(promos)
```
VivekV95, line 147: Let's keep the `hide` so that we don't have to do an empty check twice, or we can rename `hide` to `isEmpty` or something if that makes more sense

VivekV95 (changes requested):
Just one comment, also a bit confused. I might be misinterpreting but it sounds like the ticket is stating that we _dont_ want empty space when there are no promos whereas the implementation is ensuring there's empty space

app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
vespertineSiren: > Just one comment, also a bit confused. I might be misinterpreting but it sounds like the ticket is stating that we _dont_ want empty space when there are no promos whereas the implementation is ensuring there's empty space

If we remove the min height for viewMain in the layout_promo_section then we should be able to just hide completely.

VivekV95: ^ Works for me

VivekV95:
Looks like there's bottom padding in the root view which causes empty space at the bottom no matter what scroll position we're in. I'm guessing we only want that space once we've scrolled to the bottom. 

Screen Shot 2021-06-08 at 1 41 59 PM

VivekV95 (approved):

onodude (approved):
thanks!

PR #109, opened 2021-06-07
TCM-490 | Improve `Videos` screen

vespertineSiren:
### Summary

- Remove button to start workout, no related workouts label. 

### Issue

[TCM-490](https://hyfnla.atlassian.net/browse/TCM-490)

### Video/Screenshot

https://user-images.githubusercontent.com/27742105/121099272-6653c380-c7ac-11eb-875c-85611c7a37eb.mov

| | | Most Recent Changes |
|----|----|----|
|Screenshot_1623107467|Screenshot_1623107478|image|

### DIFF
--- a/app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
+++ b/app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
@@ -14,45 +14,44 @@ import com.trx.consumer.models.common.TrainerModel
 import com.trx.consumer.models.common.VideoModel
 import com.trx.consumer.models.common.VideosModel
 import com.trx.consumer.screens.discover.list.DiscoverAdapter
-import com.trx.consumer.screens.player.PlayerActivity
 
 class VideosFragment : BaseFragment(R.layout.fragment_videos) {
+
+    //region Objects
 
     private val viewModel: VideosViewModel by viewModels()
     private val viewBinding by viewBinding(FragmentVideosBinding::bind)
 
     private lateinit var adapter: DiscoverAdapter
 
+    //endregion
+
+    //region Initializers
+
     override fun bind() {
         val model = NavigationManager.shared.params(this) as VideosModel
-
-        viewModel.apply {
-            this.model = model
-
-            eventTapBack.observe(viewLifecycleOwner, handleTapBack)
-            eventLoadView.observe(viewLifecycleOwner, handleLoadView)
-            eventTapVideo.observe(viewLifecycleOwner, handleTapVideo)
-            eventTapStartWorkout.observe(viewLifecycleOwner, handleTapStartWorkout)
-            eventTapProfile.observe(viewLifecycleOwner, handleTapProfile)
-
-            doLoadView()
-        }
 
         viewBinding.apply {
             adapter = DiscoverAdapter(viewModel) { lifecycleScope }
             rvRelatedVideos.adapter = adapter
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
```diff
@@ -14,45 +14,44 @@ import com.trx.consumer.models.common.TrainerModel
...
+
+    //region Initializers
+
     override fun bind() {
         val model = NavigationManager.shared.params(this) as VideosModel
 
+        viewBinding.apply {
+            adapter = DiscoverAdapter(viewModel) { lifecycleScope }
+            rvRelatedVideos.adapter = adapter
+            btnBack.action { viewModel.doTapBack() }
```
VivekV95, line 37: I think we can connect tapping trainer profile now

app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
```diff
@@ -14,45 +14,44 @@ import com.trx.consumer.models.common.TrainerModel
...
+        }
```
VivekV95, line 38: It looks like some programs/collections have descriptions and some don't, but we're always just displaying the description of the first video. Now that we can't play the first video here, feels like it would be better to check if `VideosModel` has a description and use that if available, and use the first video's description if not.

VivekV95 (changes requested):
Just a couple things

devender197:
@VivekV95 Agreed with you suggestions. But they seems little out of scope of this task. If we are considering to do these improvement in this task. Can we please consider added a limit with Read more if description is longer that threshold value.

VivekV95:
> @VivekV95 Agreed with you suggestions. But they seems little out of scope of this task. If we are considering to do these improvement in this task. Can we please consider added a limit with Read more if description is longer than a threshold value?

Our "Improve" tasks are open ended on purpose so we can knock out miscellaneous requirements, so I think the additional changes would be in scope, but if we want to create a separate task regardless, totally fine by me.

app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
vespertineSiren: Here we go: https://github.com/hyfn/trx-consumer-android/pull/109/commits/31fbcbfe1a0b34aae0ae9cde07907b0b65ed3cc5

app/src/main/java/com/trx/consumer/screens/videos/VideosFragment.kt
vespertineSiren: profile tap: https://github.com/hyfn/trx-consumer-android/pull/109/commits/ffed789ae7ceee11d5c0dcd3a52673b1641bbe15

VivekV95 (approved):

devender197:
> @VivekV95 Agreed with you suggestions. But they seems little out of scope of this task. If we are considering to do these improvement in this task. Can we please consider added a limit with Read more if description is longer than a threshold value?

@vespertineSiren @onodude @VivekV95 So are we handling this `"Read more if description is longer than a threshold value"` request in current PR or  it is out of scope?

devender197:
A Question asked

onodude (approved):
thanks! lets create a new task for followups @vespertineSiren @VivekV95

PR #111, opened 2021-06-08
TCM-492 |  Add equipments section

devender197:
### Summary

- [x] Equipment Section added to workout screen

### Issue

[TCM-492](https://hyfnla.atlassian.net/browse/TCM-492)

### Video

https://user-images.githubusercontent.com/24737954/121410052-062d6080-c917-11eb-8d20-91b4afd1e295.mov

### DIFF
--- a/app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
+++ b/app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
@@ -1,5 +1,6 @@
 package com.trx.consumer.screens.workout
 
+import android.text.TextUtils
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.Observer
 import com.trx.consumer.R
@@ -17,6 +18,8 @@ import com.trx.consumer.models.common.WorkoutModel
 import com.trx.consumer.models.states.BookingState
 import com.trx.consumer.models.states.WorkoutViewState
 import com.trx.consumer.screens.player.PlayerActivity
+import java.lang.StringBuilder
+import java.util.Locale
 
 class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
 
@@ -94,6 +97,9 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
             viewTrainer.isHidden = false
             imgTrainerPhoto.load(model.video.trainer.profilePhoto)
             lblTrainerName.text = model.video.trainer.fullName
+
+            viewEquipment.isHidden = model.video.equipment.isEmpty()
+            lblEquipment.text = listToString(model.video.equipment)
         }
     }
 
@@ -125,5 +131,12 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
 
     override fun onBackPressed() {
         viewModel.doTapBack()
+    }
+
+    private fun listToString(list: List<String>): String {
+        val capitalizedList = list.map { it.capitalize(Locale.ROOT) }
+        val size = list.size
+        return StringBuilder(TextUtils.join(", ", capitalizedList.subList(0, size - 1)))
+            .append(" and ${capitalizedList[size - 1]}").toString()
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -126,4 +132,11 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
     override fun onBackPressed() {
         viewModel.doTapBack()
     }
+
+    private fun listToString(list: List<String>): String {
```
devender197, line 136: @onodude need suggestion on this function name and possibility as extension exit or not?

VivekV95, line 136: Couple things: 
- Needs an empty check or we'll crash
- Don't want to prepend "and" if it's just one item

Working well in other cases

VivekV95: Also do we want an oxford comma or not? Currently we're not using one

VivekV95 (changes requested):
Just a couple things

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/111/commits/458707ce45eb3979489509c914b58cc6b4f1abe6)

onodude (changes requested):
@devender197 @VivekV95 would it be possible to mimic the changes on iOS with respect to the logic: https://github.com/hyfn/trx-consumer-ios/pull/194

@onodude Made similar to iOS @onodude [Done](https://github.com/hyfn/trx-consumer-android/pull/111/commits/2868fb736f3849604ec58b7d12a315815d16a122)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
```diff
@@ -126,4 +134,13 @@ class WorkoutFragment : BaseFragment(R.layout.fragment_workout) {
...
```
VivekV95, line 138: [This](https://github.com/hyfn/trx-consumer-android/pull/111#pullrequestreview-679054052) should be possible here with same pattern, we have a `capitalized` extension we can use and I think we should be able to just call `joinToString` with a "," separator and inline everything like iOS is doing (if we don't care about adding an "and" at the end)

VivekV95 (changes requested):
Addressing @onodude 's request

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: > @devender197 @VivekV95 would it be possible to mimic the changes on iOS with respect to the logic: [hyfn/trx-consumer-ios#194](https://github.com/hyfn/trx-consumer-ios/pull/194)

@onodude Made similar to iOS @onodude [Done](https://github.com/hyfn/trx-consumer-android/pull/111/commits/2868fb736f3849604ec58b7d12a315815d16a122)

devender197:
> @devender197 @VivekV95 would it be possible to mimic the changes on iOS with respect to the logic: [hyfn/trx-consumer-ios#194](https://github.com/hyfn/trx-consumer-ios/pull/194)


@onodude Made similar to iOS @onodude [Done](https://github.com/hyfn/trx-consumer-android/pull/111/commits/2868fb736f3849604ec58b7d12a315815d16a122)

app/src/main/java/com/trx/consumer/screens/workout/WorkoutFragment.kt
devender197: addressed [here](https://github.com/hyfn/trx-consumer-android/pull/111#issuecomment-857895530)

devender197:
@onodude only difference we have with iOS is that iOS don't have line below equipment section and android have it. do we need remove it on android also?

onodude (approved):
thanks!

PR #121, opened 2021-06-18
TCM-442 | Setup firebase remote config connection

devender197:
### Summary

- [x] Setup the firebase remote config connection

### Issue

[TCM-442](https://hyfnla.atlassian.net/browse/TCM-442)

#### Firebase parameter added
Screen Shot 2021-06-17 at 6 02 12 PM

### DIFF
--- a/app/src/main/java/com/trx/consumer/core/MainModule.kt
+++ b/app/src/main/java/com/trx/consumer/core/MainModule.kt
@@ -133,7 +133,7 @@ object MainModule {
 
     @Provides
     @Singleton
-    fun provideConfigManager(): ConfigManager = ConfigManager()
+    fun provideConfigManager(cacheManager: CacheManager): ConfigManager = ConfigManager(cacheManager)
 
     @Provides
     @Singleton
--- a/app/src/main/java/com/trx/consumer/core/MainModule.kt
+++ b/app/src/main/java/com/trx/consumer/core/MainModule.kt
@@ -133,7 +133,8 @@ object MainModule {
 
     @Provides
     @Singleton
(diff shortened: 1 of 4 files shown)


onodude (approved):
thanks!

app/src/main/java/com/trx/consumer/core/MainModule.kt
```diff
@@ -133,7 +133,7 @@ object MainModule {
 
     @Provides
     @Singleton
-    fun provideConfigManager(): ConfigManager = ConfigManager()
+    fun provideConfigManager(cacheManager: CacheManager): ConfigManager = ConfigManager(cacheManager)
```
VivekV95, line 136: Exceeds character limit

VivekV95 (changes requested):
Just one minor formatting issue

app/src/main/java/com/trx/consumer/core/MainModule.kt
onodude: should we add a return statement?

VivekV95: Stylistically I prefer using `return` statements but looks like the pattern in the file seems to be just using `=`

devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/121/commits/cf957b858f06569c9f92f87541c565fd216924f2)

onodude (approved):
thanks!

VivekV95 (approved):

PR #122, opened 2021-06-18
TCM-427 | Polish settings screen for v1

devender197:
### Summary

- [x] Membership row added to screen

### Issue

[TCM-427](https://hyfnla.atlassian.net/browse/TCM-427)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/122622059-58276200-d04c-11eb-9ac5-9347bf6cf681.mov

### Screenshot
Screen Shot 2021-06-18 at 3 42 34 PM

### DIFF
--- a/app/src/main/java/com/trx/consumer/models/common/SettingsModel.kt
+++ b/app/src/main/java/com/trx/consumer/models/common/SettingsModel.kt
@@ -31,7 +31,7 @@ class SettingsModel {
 
     val titleTextSize: Int
         get() = when (type) {
-            SettingsType.SUBSCRIPTIONS -> 10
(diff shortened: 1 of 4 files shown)


onodude (changes requested):
thanks @devender197 i noticed some design were off before this task would you be able to handle; settings option lblTitle should be size 10 and bold

devender197:
> thanks @devender197 i noticed some design were off before this task would you be able to handle; settings option lblTitle should be size 10 and bold

@onodude Thank you for the review. Changes done [here](titleTextSize in case of data available is set to 10dp) and demo has been updated.

onodude (approved):
thank you @devender197 also noticed the design specs had a back button corrected that too

PR #133, opened 2021-06-23
TCM-255 | Hardcoded Live Session

vespertineSiren:
### Summary

- Refer to iOS codebase to connect to hardcoded live session. 
- Light code cleanup, needs more refactoring. 

### Issue

[TCM-255](https://hyfnla.atlassian.net/browse/TCM-255)

### Video/Screenshot

| Live Player with 2 iOS devices | Broken Landscape Mode | iOS POV - Android Landscape |
|----|----|----|
|image|image|image|


app/src/main/java/com/trx/consumer/extensions/ActivityExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.trx.consumer.extensions
+
+import android.content.pm.PackageManager
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.content.ContextCompat
+
+fun AppCompatActivity.checkLivePermission(permission: String): Boolean =
+    ContextCompat.checkSelfPermission(this, permission) != PackageManager.PERMISSION_GRANTED
```
VivekV95, line 8: This will be replaced by `PermissionManager` once complete, let's add a todo just to make sure it'll be refactored

app/src/main/AndroidManifest.xml
```diff
@@ -42,6 +40,10 @@
             android:name=".screens.video.VideoPlayerActivity"
             android:configChanges="orientation|screenSize" />
 
+        <activity
+            android:name=".screens.liveplayer.LivePlayerActivity"
+            android:configChanges="orientation|screenSize" />
```
VivekV95, line 45: I think using a separate landscape layout is a config change, so let's add `screenLayout` to the `configChanges` and it can be inflated and set in `onConfigurationChanged` using `setContentView`. That is if we _don't_ want the activity to be destroyed and recreated on rotation.

VivekV95 (dismissed):

onodude (approved):
thank you!

PR #134, opened 2021-06-23
TCM-621 | Create CommonStateButton

devender197:
### Summary

- [x] Shapable added
- [x] CommonCheckBox.kt name changed to CommonStateButton.kt

### Issue

[TCM-621](https://hyfnla.atlassian.net/browse/TCM-621)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/123019178-d35a8200-d384-11eb-8446-0fbfebd33ef6.mov


app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.util.AttributeSet
+import androidx.annotation.RequiresApi
+import androidx.appcompat.widget.AppCompatCheckBox
+import androidx.core.content.ContextCompat
+import androidx.core.content.withStyledAttributes
+import com.trx.consumer.R
+import com.trx.consumer.common.shapeable.CommonShapeable
+import com.trx.consumer.common.shapeable.CommonShapeableHandler
+
+@RequiresApi(Build.VERSION_CODES.M)
```
VivekV95, line 15: This may be an issue, our current minimum api is 21 and the app on the store is 22. We'd have to raise our minimum api to 23 to use `buttonDrawable`, need to investigate how many users we'd affect.

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+class CommonStateButton @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet?,
+    defStyleAttr: Int = 0
+) : AppCompatCheckBox(context, attrs), CommonShapeable<CommonStateButton> {
+
+    private var checkedStateDrawable: Drawable? = null
+    private var plainStateDrawable: Drawable? = null
+    private var defaultStateDrawable: Drawable? = null
```
VivekV95, line 24: We don't need to support default state, that's what `CommonImageView` is for.

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    fun onChecked(action: (isChecked: Boolean) -> Unit) {
+        setOnCheckedChangeListener { _, isChecked ->
+            setButtonDrawable()
+            action(isChecked)
+        }
+    }
+
+    fun image(drawable: Int, state: Int?) {
```
VivekV95, line 52: Related to [this](https://github.com/hyfn/trx-consumer-android/pull/134/files#r659945101)

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+        when (state) {
+            ACTIVE -> checkedStateDrawable = defaultStateDrawable
+            PLAIN -> plainStateDrawable = defaultStateDrawable
+        }
+
+        buttonDrawable = null
+        setButtonDrawable()
+    }
+
+    companion object {
```
VivekV95, line 63: Not sure we need the stats, can just use a boolean

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    fun image(drawable: Int, state: Int?) {
+        defaultStateDrawable = ContextCompat.getDrawable(context, drawable)
+        when (state) {
+            ACTIVE -> checkedStateDrawable = defaultStateDrawable
+            PLAIN -> plainStateDrawable = defaultStateDrawable
+        }
+
+        buttonDrawable = null
```
VivekV95, line 59: Is this needed?

VivekV95 (changes requested):
Looking pretty good overall with the `buttonDrawable` approach, but have to investigate whether we'll be able to raise our API level to 23 or not. Will wait for @onodude's thoughts.

It is not required now. Only related issue with api level is resolved.

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.trx.consumer.common.shapeable.CommonShapeableHandler
+
+@RequiresApi(Build.VERSION_CODES.M)
+class CommonStateButton @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet?,
+    defStyleAttr: Int = 0
+) : AppCompatCheckBox(context, attrs), CommonShapeable<CommonStateButton> {
+
+    private var checkedStateDrawable: Drawable? = null
```
VivekV95, line 22: Think we can just make these public, would be nice to have the ability to set the drawables programatically without toggling checked state

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
VivekV95: Not a major issue as @devender197 stated it's only needed for the getter

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,64 @@
...
+        context.withStyledAttributes(attrs, R.styleable.CommonStateButton) {
+            checkedStateDrawable = getDrawable(R.styleable.CommonStateButton_checkedState)
+            plainStateDrawable = getDrawable(R.styleable.CommonStateButton_plainState)
+            defaultStateDrawable = checkedStateDrawable ?: plainStateDrawable
+                    ?: CompoundButtonCompat.getButtonDrawable(this@CommonStateButton)
+            setButtonDrawable()
+        }
+    }
+
+    private fun setButtonDrawable() {
```
VivekV95, line 37: At the moment, setting a default drawable through `image` (not passing in state) won't work as we're checking for existence of plain/active drawables as source of truth here. If we're supporting a default drawable, we probably want to introduce a third state for it, store state in the view, and use state as source of truth. Checked state + default state need to be able to co-exist. From what I understand, setting a default drawable means view is no longer checkable.

devender197: I think we just need to pass null as state. or we can improve it by setting state= null as default value. In both case, it will work. 

`btnCamera.image(R.drawable.ic_btn_back_white, null)`

VivekV95: It won't work, just tested it out and confirmed. 
Screen Shot 2021-06-28 at 4 14 50 PM
The default drawable will never be set to one of the other ones because the state isn't passed in. There's also the issue of the check listener being triggered when we're using a default drawable, which would cause issues when we're disabling a button for example. With this approach we'd want to do something like this:
Screen Shot 2021-06-28 at 5 19 10 PM
Screen Shot 2021-06-28 at 5 19 24 PM

devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/134/commits/a593a5a39c7c93f27bd30418364b0e56b362b9a9)

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun onChecked(action: (isChecked: Boolean) -> Unit) {
+        setOnCheckedChangeListener { _, isChecked ->
+            setButtonDrawable()
+            action(isChecked)
+        }
+    }
+
+    fun image(drawable: Int, state: Int? = null) {
+        defaultStateDrawable = ContextCompat.getDrawable(context, drawable)
+        state?.let {
```
VivekV95, line 52: One last thing, I still think we should enable and disable the states by using `isEnabled` depending on whether a state is set. If we're using the default state, this no longer should behave like a state button and the check changed listener should not be triggered.

VivekV95 (approved):

onodude (approved):
thank you @devender197 🚀

PR #143, opened 2021-06-24
TCM-571 | Fix `ErrorAlert` dismiss

devender197:
### Summary

- [x] Dismiss called instead of Navigation dismiss in ErrorAlert

### Issue

[TCM-571](https://hyfnla.atlassian.net/browse/TCM-571)

### Video/Screenshot

None


onodude (approved):
thanks!

PR #149, opened 2021-06-25
TCM-640 | Hide trainer profile button

VivekV95:
## Description
Hide trainer profile button

### Summary

- [x] Hide trainer profile button in Videos screen

### Issue

[TCM-640](https://hyfnla.atlassian.net/browse/TCM-640)

### Video/Screenshot

image


vespertineSiren (approved):
Approved but why do we have that action if that button is always `gone`?

VivekV95:
> Approved but why do we have that action if that button is always `gone`?

Just so we can swap out the visibility and be done with it for v2

PR #145, opened 2021-06-24
TCM-568 | Add FM containers/classes to Group/Private Player

vespertineSiren:
### Summary

 - Add  handler, listener, viewmodels, injections, layout changes, entry in TestUtilityFragment to the Group and Private Player. 

### Issue

[TCM-568](https://hyfnla.atlassian.net/browse/TCM-568)

### Video/Screenshot

| Player with working End Button |
|-----|
|image|

PR #152, opened 2021-06-28
TCM-526 | Skin LivePlayer

vespertineSiren:
### Summary

- [x] Considerable cleanup of player handler that will be brought to https://github.com/hyfn/trx-consumer-android/pull/145
- [x] Major cleanup of activity that handles permissions temporarily. 
- [x] Moved major logic to handler or activity  

### Issue

[TCM-526](https://hyfnla.atlassian.net/browse/TCM-526)

### Video/Screenshot

https://user-images.githubusercontent.com/27742105/124842954-a346dd80-df45-11eb-8f7c-5badea1e6ff9.mov

| Portrait | Landscape |
|-----|-----|
|Screenshot_1625254883|Screenshot_1625254903|

**Requesting Permission - Portrait**

https://user-images.githubusercontent.com/27742105/124206667-a0a53d80-da98-11eb-8638-0397dc840b0b.mov

**Livestream - Landscape***

https://user-images.githubusercontent.com/27742105/124207117-a3546280-da99-11eb-8c35-586f79f57634.mov


onodude:
could you please add a demo once complete

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -4,116 +4,88 @@ import android.Manifest
...
 @AndroidEntryPoint
 class LivePlayerActivity : AppCompatActivity() {
 
+    //region Objects
     private val viewModel: LivePlayerViewModel by viewModels()
+    private lateinit var viewBinding: ActivityLivePlayerBinding
 
     @Inject
-    lateinit var livePlayerHandler: LivePlayerHandler
+    lateinit var handler: LivePlayerHandler
```
VivekV95, line 29: Do the handlers need to be injected? Seems like we'll only be using them in the corresponding screens and not sure we want singleton instances hanging around in memory once the user leaves the screen.

vespertineSiren: Youre right. Not sure why I injected but I can instantiate here either way

vespertineSiren: https://github.com/hyfn/trx-consumer-android/pull/152/commits/df96e92df5b4e778cbaa5f721d5e3e913d28ee2b

onodude (approved):
thank you @vespertineSiren

onodude (approved):
thanks!

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -4,117 +4,88 @@ import android.Manifest
...
 
-    private var currentId: String = ""
-    private var sendEncodings: ArrayList<Int>? = null
-    private var recvEncodings: ArrayList<Int>? = null
-    private val prefix = "Bitrate: "
+    lateinit var container: RelativeLayout
```
VivekV95, line 33: Wanna make sure we nulliify this in `onDestroy` or it'll leak

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerViewModel.kt
```diff
@@ -5,26 +5,47 @@ import androidx.lifecycle.viewModelScope
...
 class LivePlayerViewModel @ViewModelInject constructor (
-    private val backendManager: BackendManager
+    private val backendManager: BackendManager,
+    private val analyticsManager: AnalyticsManager
 ) : BaseViewModel(), LivePlayerListener {
 
     //region Objects
 
     var model: WorkoutModel? = null
+        set(value) {
```
VivekV95, line 24: This is a bit messy, is it needed? Would anything bad happen if the same model is just set again?

vespertineSiren: Yeah I felt that it was necessary so it doesnt have to do through all those backend calls again if the orientation is changed.

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -4,117 +4,88 @@ import android.Manifest
...
 import javax.inject.Inject
 
 @AndroidEntryPoint
 class LivePlayerActivity : AppCompatActivity() {
 
+    //region Objects
     private val viewModel: LivePlayerViewModel by viewModels()
+    private lateinit var viewBinding: ActivityLivePlayerBinding
 
     @Inject
```
VivekV95, line 30: Could we inject the handler into the viewmodel instead and communicate with it through the viewmodel? Currently the pattern is a bit strange because the handler has an instance of the `viewModelScope` but exists in the activity. Also kinda defeats the purpose of having a viewmodel if we're handling this much business logic in the activity anyways

VivekV95 (changes requested):
Few thoughts

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
vespertineSiren: I think that should be a followup task. Much too large to do now.

VivekV95: Yeah that's totally fine

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerViewModel.kt
VivekV95: Why would setting the model again cause backend calls?

vespertineSiren: No it but will replace the previous model that has a valid livereponsemodel here:
```
if (liveResponse.isSuccess) {
                        try {
                            val liveModel = LiveResponseModel.parse(liveResponse.responseString)
                            if (liveModel.isValidType) {
                                workout.live = liveModel
                                localMediaStarted = true
                                eventLoadVideo.postValue(workout)
                            } else {
                                eventLoadError.postValue("There was an error")
                            }
                        } catch (e: Exception) {
                            LogManager.log(e)
                        }
                    }
```

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
vespertineSiren: Here we are: https://github.com/hyfn/trx-consumer-android/pull/152/commits/d748e198f347321db2d90b4790547fbaa5c70e4a

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerViewModel.kt
VivekV95: Yeah not a fan of that pattern at all. Also fine for a later task, but feels really wrong to couple a response model with a model. We don't do that anywhere else. Whatever info we need when entering screen should just be sent as properties in the `WorkoutModel` and the live response info should be stored separately outside the `WorkoutModel`. Fine for now though.

VivekV95 (approved):

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerViewModel.kt
vespertineSiren: Yeah. I get it. LiveResponseModel should be its own variable. We can add that to the followup.

onodude (approved):
thank you!

PR #130, opened 2021-06-22
TCM-347 | Connect backend to Schedule screen: live, trainer live, user live

devender197:
### Summary

- [x] Connected with backend for live, trainer live and user live

### Issue

[TCM-347](https://hyfnla.atlassian.net/browse/TCM-347)

#### Trainer Live

https://user-images.githubusercontent.com/24737954/123000975-173d8f00-d365-11eb-9ed5-c3df9b5900d7.mov

#### Live

https://user-images.githubusercontent.com/24737954/123001015-1f95ca00-d365-11eb-83f7-c5605971c4ae.mov

PR #153, opened 2021-06-29
TCM-349 | Connect backend to Schedule screen: user virtual

devender197:
### Summary

- [x] TCM-349 Connect backend to Schedule screen: user virtual
- [x] TCM-348 Connect backend to Schedule: trainer virtual

### Issue

[TCM-349](https://hyfnla.atlassian.net/browse/TCM-349)


app/src/main/java/com/trx/consumer/models/common/CalendarModel.kt
```diff
@@ -17,4 +20,30 @@ class CalendarModel(var state: CalendarViewState = CalendarViewState.DISPLAY) {
...
+            calendar.time = startDate
+            while (calendar.time <= end) {
+                LogManager.log("${calendar.time}")
+                val date = calendar.time
+                val item = DaysModel(date).apply { this.calendarState = state }
+                selectedDate?.let { if (it.isSameDay(date)) item.state.isSelected = true }
+                list.add(item)
+                calendar.add(Calendar.DATE, 1)
+            }
+            return list
```
vespertineSiren, line 46: The logic for this is bit verbose. Consider other Kotlin specific functions to limit the amount of code since its looking a bit like java.

app/src/main/java/com/trx/consumer/models/common/ScheduleModel.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.os.Parcelable
+import com.trx.consumer.models.states.ScheduleViewState
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+class ScheduleModel(
+    val state: ScheduleViewState,
+    val key: String? = null,
+    val trainerProgram: TrainerProgramModel? = null
+) : Parcelable
```
vespertineSiren, line 12: I made a comment about this in another PR somewhere but I feel that we should probably not use nullables here.

app/src/main/java/com/trx/consumer/models/common/CalendarModel.kt
devender197: It is would be a great help. If you **specify/point** out where we can use kotlin functions.

app/src/main/java/com/trx/consumer/models/common/ScheduleModel.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/130#discussion_r668181858)

onodude (approved):
thanks further improvements will be done with: https://hyfnla.atlassian.net/browse/TCM-732

PR #157, opened 2021-07-01
TCM-672 | improve membership

devender197:
### Summary

- [x] Info Label at bottom of membership screen
- [x] Info label at the bottom of Alert dialog

### Issue

[TCM-672](https://hyfnla.atlassian.net/browse/TCM-672)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/124195598-9c205b00-da7f-11eb-8a3c-2848c4739f5b.mov

### Screen shot

| Info button membership screen | Info button alert dialog |
|---|--|
|Screenshot_1625177744|Screenshot_1625167735|

### DIFF
--- a/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
+++ b/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
@@ -23,7 +23,7 @@ class AlertModel private constructor(
     var secondaryMethod: (() -> Unit)? = null,
     @StringRes
     var secondaryTitle: Int = R.string.alert_button_title_dismiss,
-    var clearTitle: @RawValue SpannableString? = null,
+    var clearTitle: CharSequence? = null,
     var secondaryState: AlertViewState = AlertViewState.NEUTRAL,
     var backPressAction: AlertBackAction? = null
 ) : Parcelable {
@@ -62,18 +62,12 @@ class AlertModel private constructor(
         backPressAction = alertBackAction
     }
 
-    companion object : Parceler<AlertModel> {
+    companion object {
 
         fun create(title: String, message: String): AlertModel =
             AlertModel().apply {
                 this.title = title
                 this.message = message
             }
-
-        override fun AlertModel.write(dest: Parcel, flags: Int) {
-            TextUtils.writeToParcel(clearTitle, dest, flags)
-        }
-
-        override fun create(parcel: Parcel): AlertModel = AlertModel(clearTitle = TextUtils.CHAR_SEQUENCE_CREATOR.createFromParcel(parcel) as @RawValue SpannableString)
     }
 }
--- a/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
+++ b/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
@@ -1,16 +1,12 @@
 package com.trx.consumer.models.common
 
-import android.os.Parcel
 import android.os.Parcelable
 import android.text.SpannableString
-import android.text.TextUtils
 import androidx.annotation.StringRes
 import com.trx.consumer.R
 import com.trx.consumer.screens.alert.AlertBackAction
 import com.trx.consumer.screens.alert.AlertViewState
-import kotlinx.parcelize.Parceler
 import kotlinx.parcelize.Parcelize
-import kotlinx.parcelize.RawValue
 
 @Parcelize
 class AlertModel private constructor(
--- a/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
+++ b/app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
@@ -1,11 +1,16 @@
 package com.trx.consumer.models.common
 
+import android.os.Parcel
 import android.os.Parcelable
+import android.text.SpannableString
+import android.text.TextUtils
 import androidx.annotation.StringRes
 import com.trx.consumer.R
 import com.trx.consumer.screens.alert.AlertBackAction
 import com.trx.consumer.screens.alert.AlertViewState
+import kotlinx.parcelize.Parceler
 import kotlinx.parcelize.Parcelize
+import kotlinx.parcelize.RawValue
 
 @Parcelize
 class AlertModel private constructor(
@@ -18,6 +23,7 @@ class AlertModel private constructor(
     var secondaryMethod: (() -> Unit)? = null,
     @StringRes
     var secondaryTitle: Int = R.string.alert_button_title_dismiss,
+    var clearTitle: @RawValue SpannableString? = null,
     var secondaryState: AlertViewState = AlertViewState.NEUTRAL,
     var backPressAction: AlertBackAction? = null
 ) : Parcelable {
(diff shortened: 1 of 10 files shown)


onodude:
would it be possible to make [this](https://user-images.githubusercontent.com/2285343/124193655-0df6a580-da7c-11eb-9436-c30d91cafaca.png) part centered as well @devender197

devender197:
> would it be possible to make [this](https://user-images.githubusercontent.com/2285343/124193655-0df6a580-da7c-11eb-9436-c30d91cafaca.png) part centered as well @devender197

Thank you for review and I have made the changes requested and demo modified [here](https://github.com/hyfn/trx-consumer-android/pull/157/commits/d70be1face7e85358a1dac835408f36a3d72a508)

onodude (dismissed):
thank you @devender197

app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
```diff
@@ -42,16 +48,32 @@ class AlertModel private constructor(
...
-    companion object {
+    companion object : Parceler<AlertModel> {
 
         fun create(title: String, message: String): AlertModel =
             AlertModel().apply {
                 this.title = title
                 this.message = message
             }
+
+        override fun AlertModel.write(dest: Parcel, flags: Int) {
```
VivekV95, line 73: We don't need to implement all the parcelable stuff, can just store `clearTitle` as a `CharSequenece`. It will retain all of its Spannable properties.

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
```diff
@@ -26,6 +29,10 @@ enum class AlertViewState {
                 POSITIVE -> R.color.black
                 NEUTRAL -> R.color.greyLight
                 NEGATIVE -> R.color.red
+                CLEAR -> R.color.transparent
             }
         }
+
+    val fontFamily: Int
+        get() = if (CLEAR == this) R.font.atcarquette_regular else R.font.atcarquette_bold
```
VivekV95, line 37: `this == CLEAR`

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
```diff
@@ -26,6 +29,10 @@ enum class AlertViewState {
                 POSITIVE -> R.color.black
                 NEUTRAL -> R.color.greyLight
                 NEGATIVE -> R.color.red
+                CLEAR -> R.color.transparent
             }
         }
+
+    val fontFamily: Int
```
VivekV95, line 36: Let's annotate with `@get:FontRes`

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
```diff
@@ -54,7 +59,10 @@ class MembershipsFragment : BaseFragment(R.layout.fragment_memberships) {
     private val handleLoadView = Observer<List<MembershipModel>> { memberships ->
         LogManager.log("handleLoadView")
         adapter.update(memberships)
-        viewBinding.btnRestore.isHidden = memberships.isEmpty()
+        viewBinding.apply {
+            btnRestore.isHidden = memberships.isEmpty()
+            lblTerm.text(loadTerm())
```
VivekV95, line 64: Feel like we should hide the policy if no memberships come back as there's nothing to purchase

app/src/main/java/com/trx/consumer/screens/alert/AlertFragment.kt
```diff
@@ -47,12 +49,16 @@ class AlertFragment : BaseDialogFragment(R.layout.fragment_alert) {
...
                 }
                 btnSecondary.apply {
-                    isHidden = (secondaryTitle == 0)
-                    if (!isHidden) {
+                    if (secondaryState == AlertViewState.CLEAR) {
+                        isHidden = clearTitle == null
+                        clearTitle?.let { text(it) }
+                        typeface = ResourcesCompat.getFont(context, secondaryState.fontFamily)
+                    } else {
+                        isHidden = secondaryTitle == 0
```
VivekV95, line 57: Have to bring back the `isHidden` check below before setting text or we'll crash when hiding secondary button, as `0` isn't a valid string resource.

VivekV95 (changes requested):
Looking pretty good, left some thoughts

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
devender197: what's the difference here?

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
devender197: We haven't create any mechanism to support empty memberships neither empty membership message nor empty list row. So if we need to do that we should create separate task to do that which is out of scope of this task.

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
VivekV95: More conventional to place the the thing we're checking first.

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
VivekV95: Has to do with the design/behavior of an element that we're adding to the screen, so it would be in scope in my opinion. There's already precedent for it as we're doing the same thing with the restore button when memberships are empty for whatever reason. Definitely also need the empty list row, will create a ticket for that.

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/157/commits/f078dea696b94286b7e7f28bbe74e5e8a3e61c10)

app/src/main/java/com/trx/consumer/models/common/AlertModel.kt
devender197: @VivekV95 Thank you for that. You made it very simple. [Done](https://github.com/hyfn/trx-consumer-android/pull/157/commits/89bf6605bea86454f532dff68499df9a440d33b7#diff-0f9aa846babf82762ee220845605afea58458034f4a3b119f0f96ef5350b9a3eR26)

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/157/commits/89bf6605bea86454f532dff68499df9a440d33b7#diff-630d76159fb59d76ebfa7ccceaf8405e1e61174c1de05f1e2f2f10a68831ef81R65)

app/src/main/java/com/trx/consumer/screens/alert/AlertFragment.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/157/files#diff-32a7889db7bcf693fa0187c648be2b84a0c3e142f0ec6577a878f6176a470634R57)

devender197:
@onodude @VivekV95 Thank you for the review. Changes are implemented.

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
devender197: @onodude I think Footer with static info should not depend on dynamic data and we can pull this out of the scrollview. Dynamic data in center can change and we can display the static information every time regard less of dynamic data. We need to implement the message if in case, we don't have any membership.

app/src/main/java/com/trx/consumer/screens/alert/AlertViewState.kt
onodude: not a big deal but yes that is way

app/src/main/java/com/trx/consumer/screens/memberships/MembershipsFragment.kt
onodude: i did not get the whole convo here but in the end lets hide the lblTerm on initial state and show it after loading memberships

VivekV95: ^Yeah that's what I meant, and it stays hidden if memberships are empty for whatever reason. Sorry if I was confusing.

VivekV95 (approved):
Thanks for the changes!

onodude (approved):

PR #165, opened 2021-07-08
TCM-717 | Update CommonStateButton

devender197:
### Summary

- [x] ` isChecked = true` condition now able to trigger the drawable change.

### Issue

[TCM-717](https://hyfnla.atlassian.net/browse/TCM-717)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/124989885-7bfb1980-dff4-11eb-9d40-182541f3e0b8.mov

### DIFF
--- a/app/src/main/java/com/trx/consumer/screens/privateplayer/PrivatePlayerActivity.kt
+++ b/app/src/main/java/com/trx/consumer/screens/privateplayer/PrivatePlayerActivity.kt
@@ -24,7 +24,6 @@ class PrivatePlayerActivity : AppCompatActivity() {
         setContentView(binding.root)
 
         binding.apply {
-            btnCamera.isChecked = true
             btnCamera.onChecked { isChecked -> handleTapCamera(isChecked) }
             btnClock.onChecked { isChecked -> handleTapClock(isChecked) }
             btnMicrophone.onChecked { isChecked -> handleTapMicrophone(isChecked) }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/trx/consumer/screens/privateplayer/PrivatePlayerActivity.kt
```diff
@@ -24,6 +24,7 @@ class PrivatePlayerActivity : AppCompatActivity() {
         setContentView(binding.root)
 
         binding.apply {
+            btnCamera.isChecked = true
```
vespertineSiren, line 27: Shouldn't we set the default state in the so there's less code in the activity to deal with?

vespertineSiren (approved):
Changes make sense to me.

app/src/main/java/com/trx/consumer/screens/privateplayer/PrivatePlayerActivity.kt
onodude: i would agree with this, lets do it if possible

onodude (changes requested):
@devender197 a question or an issue :)

app/src/main/java/com/trx/consumer/screens/privateplayer/PrivatePlayerActivity.kt
devender197: That code was for testing. I am sorry. I removed it.
[Done](https://github.com/hyfn/trx-consumer-android/pull/165/commits/cbe5d3f79cce4217deb4e79bd7c43752841ef1ac)

onodude (approved):
thanks!

PR #166, opened 2021-07-09
TCM-725 | Improve touches for CommonStateButton

devender197:
### Summary

- [x] Touch feature improved for CommonStateButton

### Issue

[TCM-725](https://hyfnla.atlassian.net/browse/TCM-725)

### Video/Screenshot
https://user-images.githubusercontent.com/24737954/125126360-d6a57b80-e0af-11eb-930e-60f322064528.mov

### ScreenShot
Screen Shot 2021-07-09 at 12 09 34 PM

### DIFF
--- a/app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
+++ b/app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
@@ -1,58 +1,61 @@
 package com.trx.consumer.common
 
 import android.content.Context
-import android.graphics.drawable.Drawable
 import android.util.AttributeSet
-import android.view.View
-import androidx.core.content.ContextCompat
+import androidx.appcompat.widget.AppCompatImageButton
+import androidx.core.content.res.getResourceIdOrThrow
 import androidx.core.content.withStyledAttributes
 import com.trx.consumer.R
 import com.trx.consumer.common.shapeable.CommonShapeable
 import com.trx.consumer.common.shapeable.CommonShapeableHandler
-import com.trx.consumer.extensions.action
+import com.trx.consumer.extensions.color
+import com.trx.consumer.managers.LogManager
 
 class CommonStateButton @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet?,
     defStyleAttr: Int = 0
-) : View(context, attrs, defStyleAttr), CommonShapeable<CommonStateButton> {
+) : AppCompatImageButton(context, attrs, defStyleAttr), CommonShapeable<CommonStateButton> {
 
-    private var checkedStateDrawable: Drawable? = null
-    private var plainStateDrawable: Drawable? = null
+    private var checkedStateDrawable: Int = 0
+    private var plainStateDrawable: Int = 0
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/attrs.xml
```diff
@@ -47,6 +47,8 @@
     <declare-styleable name="CommonStateButton">
         <attr name="checkedState" format="reference"/>
         <attr name="plainState" format="reference"/>
+        <attr name="isChecked" format="boolean"/>
+        <attr name="bgColor" format="reference"/>
```
VivekV95, line 51: Don't think we need the extra `bgColor` attr here, `CommonShapeable` already accepts a `backgroundColor` which is handled when `shapeableHandler.init` is called.

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
```diff
@@ -1,60 +1,59 @@
...
         context.withStyledAttributes(attrs, R.styleable.CommonStateButton) {
-            checkedStateDrawable = getDrawable(R.styleable.CommonStateButton_checkedState)
-            plainStateDrawable = getDrawable(R.styleable.CommonStateButton_plainState)
+            isChecked = getBoolean(R.styleable.CommonStateButton_checkedState, false)
+            checkedStateDrawable = getResourceId(R.styleable.CommonStateButton_checkedState, 0)
+            plainStateDrawable = getResourceId(R.styleable.CommonStateButton_plainState, 0)
             setButtonDrawable()
-            onChecked(null)
+
+            val bgColor = getResourceId(R.styleable.CommonStateButton_bgColor, 0)
```
VivekV95, line 37: Related to [this](https://github.com/hyfn/trx-consumer-android/pull/166/files#r667170791)

VivekV95 (changes requested):
Just one thing, looking good otherwise

app/src/main/res/values/attrs.xml
devender197: [Removed](https://github.com/hyfn/trx-consumer-android/pull/166/commits/14b1f845709c5ebd4cae917892d6e24b64be795a)

app/src/main/java/com/trx/consumer/common/CommonStateButton.kt
devender197: [Removed](https://github.com/hyfn/trx-consumer-android/pull/166/commits/14b1f845709c5ebd4cae917892d6e24b64be795a)

VivekV95:
Actually, we we still have the issue of images that fill the view cropping the border. This won't be an issue with favorites or player buttons so I'm fine with merging this approach for now, but still not an ideal long-term solution. Still think just wrapping an image button in a `CommonView` is the way to go, because `CommonView` shouldn't allow child views to crop its borders (hopefully).
Screen Shot 2021-07-09 at 2 47 58 PM

VivekV95 (approved):

onodude (approved):
thank you!

PR #170, opened 2021-07-12
TCM-734 | Add UserId for Amplitude 

vespertineSiren:
### Summary

- Inject analytics manager to backendmanager, call `AnalyticsManager.setUserId(id: String)` in `BackendManager.user()`.

### Issue

[TCM-734](https://hyfnla.atlassian.net/browse/TCM-734)

### DIFF
--- a/app/src/main/java/com/trx/consumer/core/MainModule.kt
+++ b/app/src/main/java/com/trx/consumer/core/MainModule.kt
@@ -122,8 +122,17 @@ object MainModule {
 
     @Provides
     @Singleton
-    fun provideBackendManager(api: BaseApi, cacheManager: CacheManager): BackendManager =
-        BackendManager(api, cacheManager)
+    fun provideAnalyticsManager(configManager: ConfigManager): AnalyticsManager =
+        AnalyticsManager(configManager)
+
+    @Provides
+    @Singleton
+    fun provideBackendManager(
+        api: BaseApi,
+        cacheManager: CacheManager,
+        analyticsManager: AnalyticsManager
+    ): BackendManager =
+        BackendManager(api, cacheManager, analyticsManager)
 
     @Provides
     @Singleton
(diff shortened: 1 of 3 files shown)


dkhamsing:
@vespertineSiren in ios we set the amplitude user id in BackendManager.user (similar idea to xpass where it is done in BackendManager.profile).. is that something we can do in android

app/src/main/java/com/trx/consumer/screens/home/HomeViewModel.kt
```diff
@@ -60,6 +60,7 @@ class HomeViewModel @ViewModelInject constructor(
             backendManager.user()
             eventShowHud.postValue(false)
             cacheManager.user()?.let { user ->
+                analyticsManager.setUserId(user.uid)
```
onodude, line 63: shouldn't this be in the backend manager where we do: `IAPManager.shared.identify(user.uid)`

onodude (changes requested):
there are lots of changes?

app/src/main/java/com/trx/consumer/screens/home/HomeViewModel.kt
vespertineSiren: Yeah just clarified it with @dkhamsing over slack. Will revert.

vespertineSiren:
> there are lots of changes?

A lot of them are ktlint related. Will revert.

app/src/main/java/com/trx/consumer/screens/home/HomeViewModel.kt
vespertineSiren: Here you go: https://github.com/hyfn/trx-consumer-android/pull/170/commits/e40226c4256ff732a898b71d80af4e92a2213998

onodude (approved):
thanks!

PR #178, opened 2021-07-22
TCM-759 | Video Workout Placeholder

vespertineSiren:
### Summary

- [x] Remove placeholder video workout. 
- [x] Run ktlintformatting to handle missed cleaning

### Issue

[TCM-759](https://hyfnla.atlassian.net/browse/TCM-759)

### Video/Screenshot

| Home | Workouts |
|----|----|
|image|image|

### DIFF
--- a/app/src/main/java/com/trx/consumer/extensions/CommonImageViewExtensions.kt
+++ b/app/src/main/java/com/trx/consumer/extensions/CommonImageViewExtensions.kt
@@ -12,18 +12,24 @@ import com.trx.consumer.R
 import com.trx.consumer.common.CommonImageView
 
 fun CommonImageView.load(
-    url: String
+    url: String,
+    usePlaceholder: Boolean = true
 ) {
     Glide.with(context)
         .load(url)
         .fitCenter()
-        .apply(
-            RequestOptions().apply {
-                placeholder(R.drawable.ic_img_placeholder)
-                fallback(R.drawable.ic_img_placeholder)
+        .let { requestBuilder ->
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/trx/consumer/extensions/CommonImageViewExtensions.kt
```diff
@@ -12,18 +12,24 @@ import com.trx.consumer.R
 import com.trx.consumer.common.CommonImageView
 
 fun CommonImageView.load(
-    url: String
+    url: String,
+    usePlaceholder: Boolean = true
```
onodude, line 16: lets do something like:
`load(url: String, placeholder: Int? = nil)`
and create it as a new method, in the end when we call `img.load(imgUrl, nil)` should not be using a placeholder

onodude (changes requested):
thanks @vespertineSiren added a comment

app/src/main/java/com/trx/consumer/screens/discover/DiscoverViewModel.kt
```diff
@@ -135,7 +134,7 @@ class DiscoverViewModel @ViewModelInject constructor(
     }
 
     fun doTrackPageView() {
-        analyticsManager.trackPageView(AnalyticsPageModel.DISCOVER)
+        analyticsManager.trackPageView(DISCOVER)
```
onodude, line 137: thanks!

onodude (approved):
thanks @vespertineSiren

PR #203, opened 2021-08-05
TCM-789 | Connect LivePlayer rotate button

devender197:
### Summary

- [ ] User can go from portrait to landscape mode either by pressing rotate button or rotating device
- [ ] User cannot go from landscape to portrait by rotating device, can only do so by pressing rotate button
- [ ] When going from landscape to portrait by pressing rotate button, user should be given time to rotate device before the device auto-rotates back to landscape
- [ ] User can automatically go from landscape to reverse landscape by rotating 180 degrees, but should not go to portrait in between

### Issue

[TCM-789](https://hyfnla.atlassian.net/browse/TCM-789)

### Video/Screenshot

https://user-images.githubusercontent.com/24737954/128280949-d5d32ced-c290-4891-9815-3b0d04a46249.mov


app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -246,5 +262,24 @@ class LivePlayerActivity : AppCompatActivity() {
             )
         }
     }
+
+    private fun rotate() {
+        orientationJob = CoroutineScope(Dispatchers.IO).launch {
+            delay(timeMillis = 500)
```
VivekV95, line 268: Let's give the user 3 seconds like in video player, half a second is going to be too quick

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -28,6 +36,7 @@ class LivePlayerActivity : AppCompatActivity() {
     //region Objects
     private val viewModel: LivePlayerViewModel by viewModels()
     private lateinit var viewBinding: ActivityLivePlayerBinding
+    private var orientationJob: Job? = null
```
VivekV95, line 39: We have access to `lifecycleScope` here so don't think we need to manage the job

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -246,5 +262,24 @@ class LivePlayerActivity : AppCompatActivity() {
             )
         }
     }
+
+    private fun rotate() {
+        orientationJob = CoroutineScope(Dispatchers.IO).launch {
```
VivekV95, line 267: Related to [this](https://github.com/hyfn/trx-consumer-android/pull/203/files#r683685557)

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
VivekV95: Also this isn't where we need the delay and coroutine scope are needed, these are needed _after_ rotation occurs (up where landscape logic is handled in `bind`)

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -246,5 +262,24 @@ class LivePlayerActivity : AppCompatActivity() {
...
+            delay(timeMillis = 500)
+            withContext(Dispatchers.Main) {
+                if (requestedOrientation == ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE ||
```
VivekV95, line 270: Can use the same check being used in `bind` instead of having to check for landscape/user landscape/sensor landscape etc.

VivekV95 (changes requested):
Few thoughts on the first pass, let's try to keep logic similar to video player. Can elaborate on any of the comments, please feel free to reach out.

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
devender197: @onodude I don't this delay part of task even needed. It is creating a weird experience to user. If we still want to apply this it should be as minimum as possible. @VivekV95 Please [this](https://github.com/hyfn/trx-consumer-android/pull/203#discussion_r683693626) need more idea.

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/203/commits/73fc554188637b2095eeb2440f4b0ff8692ef19f)

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/203/commits/73fc554188637b2095eeb2440f4b0ff8692ef19f)

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
devender197: [Done](https://github.com/hyfn/trx-consumer-android/pull/203/commits/73fc554188637b2095eeb2440f4b0ff8692ef19f)

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
devender197: [Done]( https://github.com/hyfn/trx-consumer-android/pull/203/commits/73fc554188637b2095eeb2440f4b0ff8692ef19f)

VivekV95 (approved):
Thanks for the changes!

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
```diff
@@ -246,5 +259,29 @@ class LivePlayerActivity : AppCompatActivity() {
             )
         }
     }
+
+    private fun rotate() {
```
vespertineSiren, line 263: can we change the function to something like `doTapRotate` or `handleTapRotate` or something to that effect? It sounds a little out of place with the rest of the function names in this activity.

vespertineSiren (changes requested):
Just one comment.  seems to work ok on emulator. Will need testing on physical device.

app/src/main/java/com/trx/consumer/screens/liveplayer/LivePlayerActivity.kt
VivekV95: Maybe we could wrap it in an event

PR #213, opened 2021-08-06
TCM-827 | Improve `Home` screen

devender197:
### Summary

- [ ] Active membership added to home screen below user name

### Issue

[TCM-827](https://hyfnla.atlassian.net/browse/TCM-827)

### Video/Screenshot
|Before|After|
|--|--|
|Screenshot_1628275313|Screenshot_1628274779|


app/src/main/java/com/trx/consumer/screens/home/HomeFragment.kt
```diff
@@ -158,6 +159,10 @@ class HomeFragment : BaseFragment(R.layout.fragment_home) {
         viewBinding.apply {
             with(requireContext()) {
                 lblUserName.text = getString(R.string.home_user_name_label, model.firstName)
+                model.activeMemberships.keys.firstOrNull()?.let { membership ->
+                    lblUserMembership.text =
+                        lblUserMembership.context.spannableString(membership, fullUnderline = true)
```
VivekV95, line 164: Can just call `spannableString` directly as we're already in the scope of the context

VivekV95 (changes requested):
Just one thing

